PR #44, opened 2019-04-24
fix: Remove package dev deps and bump root dev deps.

milesj:

## Description

I noticed that anytime core is released, all packages that rely on core as a dev dependency also get bumped: https://github.com/airbnb/lunar/releases 

Lerna's logic isn't smart enough to avoid this at the moment, so let's just remove the dev deps (build still works).

Also bumped root dev deps while I was at it.

## Motivation and Context

Stop needlessly releasing packages.

## Testing

Local testing.

## Checklist


packages/forms/src/index.ts
```diff
@@ -16,7 +16,6 @@ import TextArea from './components/Form/TextArea';
 import ToggleButtonController from './components/Form/ToggleButtonController';
 import connectToForm from './composers/connectToForm';
 
-// eslint-disable-next-line import/export
```
milesj: Fixed in latest version.

PR #49, opened 2019-04-26
new(MessageItem): Add disable title translation prop.

raimondeee:

## Description

Adding a new prop that wraps the title in a span that disables translation option from Google.

## Checklist


packages/core/src/components/MessageItem/index.tsx
```diff
@@ -122,6 +126,12 @@ export class MessageItem extends React.Component<Props & WithStylesProps> {
       warning && styles.container_warning,
     );
 
+    const formatedTitle = disableTitleTranslation ? (
```
milesj: @stefhatcher Any reason `title` doesn't support nodes? Seems a bit odd to hardcode this.

milesj:
@raimondeee You need another commit or otherwise the release will fail.

raimondeee:
> @raimondeee You need another commit or otherwise the release will fail.

In addition to merging to the master? Or was that resolved by merging?

milesj:
@raimondeee You need another commit for the merge to work correctly.

packages/core/src/components/MessageItem/index.tsx
stefhatcher: it's used in other places that are string only

milesj: 👍

raimondeee:
Added another commit @milesj and merged to master - please let me know if there are any corrections or additional steps I need to take. Thank you both!

milesj:
Looks good, merging!

PR #52, opened 2019-04-29
Adding optional size property for checkbox

MarkKahn:

## Description

Adds an optional size prop to checkboxes

## Motivation and Context

Need a larger checkbox per design

## Checklist


milesj:
Do you have a screenshot of the design? This seems very arbitrary and unnecessary.

stefhatcher:
Agreed, this seems unnecessary. To keep the design system consistent, I would vote no on this approach. If design would like a larger checkbox, we should also add large radios and inputs that scale the same way as large buttons for consistency. Or if design is heading towards a future iteration where checkboxes and radios are larger, we'd make that the new default.

MarkKahn:
Screen Shot 2019-05-03 at 11 44 33 AM

Thoughts on how else to accomplish this then?

milesj:
Closing this for now.

PR #55, opened 2019-05-02
update(Query,Mutation): Limit error message to network errors

MarkKahn:

## Description

Add a prop to make `<ErrorMessage />` only show up on network errors.  Any graphql errors will pass through `null` values along with the error object.

## Motivation and Context

There are two types of possible errors: Network errors & graphql errors.  We want to pass through graphql errors so that client can a) decide how to display them and b) show the data for the part of the query that succeeded which is often the majority of it.

[Apollo docs](https://www.apollographql.com/docs/react/features/error-handling#options)

## Testing

Created forced errors and verified that this passes through gql errors and still fails with disconnected wifi

## Checklist

### DIFF
--- a/packages/apollo/src/components/Query/index.tsx
+++ b/packages/apollo/src/components/Query/index.tsx
@@ -47,7 +47,7 @@ export default class Query<Data = any, Vars = OperationVariables> extends React.
       return renderElementOrFunction(this.props.loading) || <Loader static />;
     }
 
-    if (result.error) {
+    if (result.error && result.error.networkError) {
       return (
         renderElementOrFunction(this.props.error, result.error) || (
           <ErrorMessage error={result.error} />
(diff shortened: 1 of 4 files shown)


packages/apollo/src/components/Query/index.tsx
```diff
@@ -47,7 +47,7 @@ export default class Query<Data = any, Vars = OperationVariables> extends React.
       return renderElementOrFunction(this.props.loading) || <Loader static />;
     }
 
-    if (result.error) {
+    if (result.error && result.error.networkError) {
       return (
         renderElementOrFunction(this.props.error, result.error) || (
```
milesj: Need to update `renderElementOrFunction` and `ErrorMessage` to accept `result.error.networkError` instead of `result.error`.

milesj:
@MarkKahn Is this still an issue?

MarkKahn:
yes, will update it

MarkKahn:
@milesj - PTAL

packages/apollo/test/components/Mutation.test.tsx
```diff
@@ -37,7 +37,11 @@ describe('Mutation', () => {
       },
       result: {
         data: {
-          updateSomething: {},
+          updateSomething: {
```
MarkKahn: FYI test was failing because this didn't return data.  No idea why it failed below and not here, but it was a promise failing so maybe just jest catching it in the wrong place?

milesj: Thanks for fixing 🙏

PR #60, opened 2019-05-15
update(AppLoader): Add errorTitle prop

MarkKahn:

The `<AppLoader />` component currently only shows 'Unknown Error' or `error.error_code` as the error title.  This allows an error title to be passed through to the `<ErrorMessage />` component

## Motivation and Context

Need a better title than 'Unknown Error'

## Screenshots

old:

Screen Shot 2019-05-15 at 12 05 26 PM

new:

Screen Shot 2019-05-15 at 12 05 43 PM

## Checklist


milesj:
I'd rather update `ErrorMessage` to inherit the title from `Error`, otherwise this approach is brittle. What happens when the error isn't a 404 but the `errorTitle` is hard-coded to "Not Found"?

MarkKahn:
Adjusting `ErrorMessage` to read `title` is a good idea, but I still think this should be implemented as well, otherwise you can't just pass through an actual `Error` without it saying `Unknown Error`.  `Page not found` may not have been the best example

PR #65, opened 2019-05-21
fix: Remove Babel runtime setup

milesj:

## Description

I initially setup Lunar to build with Babel runtime + CoreJS v3 for smaller file sizes, but this seems to cause many issues downstream, as using CoreJS v2 and CoreJS v3 in parallel can be quite problematic. It's also complicated to configure correctly (as seen by the storybook config).

While removing the runtime setup, I logged the before and after state of the entire "esm" folder for each package (kilobytes in the table below). While the runtime did have smaller file sizes, it wasn't that much savings in the grand scheme of things. 

| | With runtime | Without |
| ---  | --- | --- |
| apollo | 9,462 | 10,500 |
| app-shell | 8,855 | 9,564 |
| core | 545,207 | 572,108 |
| forms | 40,679 | 39,455 |
| icons | 98,953 | 139,280 |
| layouts | 10,673 | 13,356 |
| metrics | 3,928 | 3,671 |
| test-utils | 3,997 | 3,997 |

Once the community has adopted CoreJS v3 more, we can revisit this feature.

## Motivation and Context

Make downstream consumption much easier.

## Testing

Tests and lints pass. Storybook still renders correctly.

## Screenshots

N/A

## Checklist

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -12,13 +12,6 @@ const plugins = [
   '@babel/plugin-proposal-class-properties',
   '@babel/plugin-proposal-optional-catch-binding',
   '@babel/plugin-syntax-dynamic-import',
-  [
-    '@babel/plugin-transform-runtime',
-    {
-      corejs: 3,
-      regenerator: false,
-    },
-  ],
 ];
 
 if (process.env.NODE_ENV === 'test') {
--- a/babel.config.js
+++ b/babel.config.js
@@ -40,11 +40,11 @@ module.exports = {
     [
       '@babel/preset-env',
       {
-        corejs: 3,
+        // corejs: 3,
         loose: true,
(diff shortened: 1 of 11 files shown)


babel.config.js
```diff
@@ -47,11 +40,11 @@ module.exports = {
     [
       '@babel/preset-env',
       {
-        corejs: 3,
+        // corejs: 3,
         loose: true,
         modules: ESM ? false : 'commonjs',
         shippedProposals: true,
-        useBuiltIns: 'usage',
+        // useBuiltIns: 'usage',
```
stefhatcher: still want these around?

milesj: Yeah, for when I re-enable in the futurez.

PR #66, opened 2019-05-22
update(Chip): Add new compact and active props

schillerk:
to: @milesj @stefhatcher @hayes @alecklandgraf @kenchendesign 

## Description
Is there any reason we don't use `defaultProps` for `Chip`?

Right now, I just use the existing hover state, but Ken's original design uses border color instead of background. I don't know how to make something like this configurable through props since `withStyles` doesn't have access to props and inline styles can't describe hover effects.

## Motivation and Context
See previous discussion here:
https://git.musta.ch/airbnb/solar/issues/824

## Testing
Tested in storybook.

## Screenshots
Screen Shot 2019-05-22 at 4 19 29 PM

## Checklist

### DIFF
--- a/packages/core/src/components/Chip/index.tsx
+++ b/packages/core/src/components/Chip/index.tsx
@@ -7,8 +7,12 @@ import ProfilePhoto from '../ProfilePhoto';
 import ButtonOrLink, { ButtonOrLinkTypes } from '../private/ButtonOrLink';
 
 export type Props = {
+  /** Renders with a primary background and white text */
+  active: boolean;
   /** Primary chip contents. */
   children: NonNullable<React.ReactNode>;
+  /** Renders with less padding and sharper corners */
+  compact: boolean;
   /** Disabled / gray. */
   disabled?: boolean;
   /** Icon to render to the right of the primary content. */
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Chip/index.tsx
```diff
@@ -7,8 +7,12 @@ import ProfilePhoto from '../ProfilePhoto';
 import ButtonOrLink, { ButtonOrLinkTypes } from '../private/ButtonOrLink';
 
 export type Props = {
+  /** Renders with a primary background and white text */
```
milesj: Nit, but could you add a period to the end of these sentences.

packages/core/src/components/Chip/index.tsx
```diff
@@ -107,6 +124,26 @@ export default withStyles(({ color, font, pattern, transition, ui, unit }) => ({
...
+  chip_active: {
+    background: color.core.primary[3],
+    borderColor: color.core.primary[3],
+    color: color.accent.bg,
+  },
+
+  chip_active_button: {
+    '@selectors': {
+      ':not([disabled]):hover': {
+        backgroundColor: color.core.primary[2],
```
milesj: Hovers should always be darker, so use `[4]`.

packages/core/src/components/Chip/index.tsx
milesj: There were 2 of them :P

schillerk: 😭

PR #67, opened 2019-05-23
break(Chip): Add beforeIcon and afterIcon props. Remove icon prop.

schillerk:

## Description
Adds a prop for `before` to `Chip`, allowing developers to pass in an `ReactNode` to render on the left side of the `Chip`.

We could remove `profileImageSrc` since it's now possible to pass in a `ProfilePhoto` directly, but this would be a breaking change. For now, I'm just using `mutuallyExclusiveTrueProps` since passing in both would break formatting.

Could also make sense to rename `icon` to `after`, or even just implement the whole thing as a wrapped `Row`, but again, all breaking changes we'll probably want to avoid for now.

## Motivation and Context
https://git.musta.ch/airbnb/solar/issues/826

## Testing
Tested in storybook.

## Screenshots
Screen Shot 2019-05-22 at 5 26 38 PM

## Checklist

### DIFF
--- a/packages/core/src/components/Chip.story.tsx
+++ b/packages/core/src/components/Chip.story.tsx
@@ -35,7 +35,7 @@ storiesOf('Core/Chip', module)
       </Spacing>
 
       <Spacing right={0} inline>
-        <Chip beforeIcon={<IconCalendar size="1.2em" />}>Calendar</Chip>
+        <Chip before={<IconCalendar size="1.2em" />}>Calendar</Chip>
       </Spacing>
     </>
   ))
--- a/packages/core/src/components/Chip.story.tsx
+++ b/packages/core/src/components/Chip.story.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
+import IconCalendar from '@airbnb/lunar-icons/lib/general/IconCalendar';
 import IconCloseAlt from '@airbnb/lunar-icons/lib/interface/IconCloseAlt';
 import IconSettings from '@airbnb/lunar-icons/lib/interface/IconSettings';
 import lunar from ':storybook/images/lunar-logo.png';
@@ -27,7 +28,17 @@ storiesOf('Core/Chip', module)
       </Spacing>
     </>
   ))
-  .add('With an icon.', () => <Chip icon={<IconCloseAlt decorative size="2em" />}>Chip</Chip>)
+  .add('With a icons before or after.', () => (
+    <>
+      <Spacing right={1} inline>
+        <Chip icon={<IconCloseAlt size="2em" />}>Chip</Chip>
+      </Spacing>
+
+      <Spacing right={0} inline>
+        <Chip beforeIcon={<IconCalendar size="1.2em" />}>Calendar</Chip>
+      </Spacing>
+    </>
+  ))
   .add('With an icon button.', () => (
     <>
       <Spacing right={1} inline>
--- a/packages/core/src/components/Chip.story.tsx
+++ b/packages/core/src/components/Chip.story.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
 import IconCalendar from '@airbnb/lunar-icons/lib/general/IconCalendar';
+import IconUser from '@airbnb/lunar-icons/lib/general/IconUser';
 import IconCloseAlt from '@airbnb/lunar-icons/lib/interface/IconCloseAlt';
 import IconSettings from '@airbnb/lunar-icons/lib/interface/IconSettings';
 import lunar from ':storybook/images/lunar-logo.png';
@@ -13,10 +14,57 @@ storiesOf('Core/Chip', module)
     inspectComponents: [Chip],
   })
   .add('Simple chip.', () => <Chip>Chip</Chip>)
+  .add('With a icons before or after.', () => (
+    <>
+      <Spacing right={1} inline>
+        <Chip icon={<IconCloseAlt size="2em" />}>Chip</Chip>
+      </Spacing>
+
+      <Spacing right={1} inline>
+        <Chip beforeIcon={<IconCalendar size="1.2em" />}>Calendar</Chip>
+      </Spacing>
+
+      <Spacing right={1} inline>
+        <Chip icon={<IconCloseAlt size="2em" />} beforeIcon={<IconUser size="1.2em" />}>
+          Calendar
+        </Chip>
+      </Spacing>
+
+      <Spacing right={0} inline>
+        <Chip icon={<IconCloseAlt size="2em" />} beforeIcon={<IconCalendar size="1.2em" />}>
+          Calendar
+        </Chip>
+      </Spacing>
+    </>
+  ))
   .add('Compact chip.', () => (
     <>
       <Spacing right={1} inline>
-        <Chip compact onClick={action('onClick')}>
+        <Chip profileImageSrc={lunar} compact onClick={action('onClick')}>
+          Chip
+        </Chip>
+      </Spacing>
+
+      <Spacing right={1} inline>
+        <Chip beforeIcon={<IconUser size="1.2em" />} compact onClick={action('onClick')}>
+          Chip
+        </Chip>
+      </Spacing>
+
+      <Spacing right={1} inline>
+        <Chip afterIcon={<IconCalendar size="1.2em" />} compact onClick={action('onClick')}>
+          Chip
+        </Chip>
+      </Spacing>
+
+      <Spacing right={1} inline>
+        <Chip
+          active
+          beforeIcon={<IconUser size="1.2em" />}
+          afterIcon={<IconCalendar size="1.2em" />}
+          compact
+          onClick={action('onClick')}
+        >
           Chip
         </Chip>
       </Spacing>
@@ -25,17 +73,6 @@ storiesOf('Core/Chip', module)
         <Chip compact active onClick={action('onClick')}>
           Chip
         </Chip>
-      </Spacing>
-    </>
-  ))
-  .add('With a icons before or after.', () => (
-    <>
-      <Spacing right={1} inline>
-        <Chip icon={<IconCloseAlt size="2em" />}>Chip</Chip>
-      </Spacing>
-
-      <Spacing right={0} inline>
-        <Chip beforeIcon={<IconCalendar size="1.2em" />}>Calendar</Chip>
       </Spacing>
     </>
   ))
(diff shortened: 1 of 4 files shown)


milesj:
@schillerk My thoughts. Let's rename `before` to `beforeIcon` and add an `afterIcon` prop as well, which will replace `icon`. They both will be supported at the moment, but we should log a deprecation message when `icon` is used, and then we'll remove in the next major version.

schillerk:
Added after icon, not sure how to do the prop-types logic for icon and afterIcon. Should be something like:

- `onIconClick` requires `icon` or `afterIcon`
- `icon` and `afterIcon` are mutually exclusive

But I didn't see anything like requiresOneOf in airbnb-prop-types
https://github.com/airbnb/prop-types

I don't know if there's any use case for this, in theory we might one day want to turn `onIconClick` into `onBeforeIconClick` and `onAfterIconClick`.

Alternatively, we could let users pass in the button wrapper directly and handle click events themselves, but then we lose being able to propagate the `disabled` to the `before`/`after` icons. There's probably way to do this where `disabled` instead renders a dark overlay that prevents click evens from passing through, but again, probably not worth doing any of this at the moment.

packages/core/src/components/Chip/index.tsx
```diff
@@ -77,34 +81,36 @@ export class Chip extends React.Component<Props & WithStylesProps> {
         )}
         {...props}
       >
-        {(before || profileImageSrc) && (
+        {(beforeIcon || profileImageSrc) && (
           <div {...css(styles.chipItem, styles.sideContent)}>
             <div {...css(styles.sideContentInner)}>
```
milesj: This should have `styles.iconWrapper` when the before icon is being used.

packages/core/src/components/Chip/index.tsx
```diff
@@ -77,34 +81,36 @@ export class Chip extends React.Component<Props & WithStylesProps> {
...
-                  disabled={disabled}
-                  onClick={onIconClick}
-                >
-                  {icon}
-                </ButtonOrLink>
-              ) : (
-                icon
-              )}
+        {afterIcon ||
+          (icon && (
```
milesj: Condition here looks wrong.

milesj: Maybe just set it above as a variable? `const afterIcon = props.afterIcon || props.icon`.

packages/core/src/components/Chip/index.tsx
```diff
@@ -41,7 +44,8 @@ export class Chip extends React.Component<Props & WithStylesProps> {
   render() {
```
milesj: Add this in the function.

```
if (__DEV__) {
  if (icon) {
    console.log('Chip: `icon` prop is deprecated, please use `afterIcon` instead.');
  }
}
```

stefhatcher:
@schillerk can you add a story example of both `beforeIcon` and `afterIcon` being used? maybe `IconUser` for `beforeIcon` and a close for `afterIcon`?

schillerk:
@stefhatcher good catch:
Screen Shot 2019-05-23 at 1 57 05 PM
Screen Shot 2019-05-23 at 1 56 48 PM

2 wonky cases:
1. User profile on compact chips
2. Default icon color on active chips

I can make `compact` and `profileImageSrc` mutually exclusive

Less sure what to do about the icon color. We could let Chip override icon color when `active` but this feels a little heavy handed. Since Chip doesn't manage its own state, maybe it's reasonable to have icon color controlled externally as well.

stefhatcher:
it looks like there's a greater space between before & chip content than chip content & after

packages/core/src/components/Chip.story.tsx
```diff
@@ -12,10 +14,51 @@ storiesOf('Core/Chip', module)
...
+      </Spacing>
+
+      <Spacing right={1} inline>
+        <Chip icon={<IconCloseAlt size="2em" />} beforeIcon={<IconUser size="1.2em" />}>
+          Calendar
+        </Chip>
+      </Spacing>
+
+      <Spacing right={0} inline>
+        <Chip icon={<IconCloseAlt size="2em" />} beforeIcon={<IconCalendar size="1.2em" />}>
```
stefhatcher: should these examples use `afterIcon` to encourage that usage instead of `icon`?

stefhatcher: aaah or can both be supported? looks like both

schillerk: Yeah, temporarily trying to preserve backwards compatibility until the next major release :(
You're right though, `afterIcon` is preferred, switching everything over, thanks!

schillerk:
@stefhatcher Good points about icon padding, I fixed before/after padding inconsistencies and found a related bug for applying the `chip_noBefore`/`chip_noAfter` styles.

Screen Shot 2019-05-23 at 3 42 35 PM
Screen Shot 2019-05-23 at 3 42 39 PM


You'll notice there doesn't seem to be a lot of padding in non-compact modes, certainly not as much as specified in the mocks:
https://media.git.musta.ch/user/905/files/a2d83ad6-7bc1-11e9-8bb9-49bce226641b

We're currently using `4px` (`unit * .05`) padding on icons, we can push it up to `8px`, but we'll lose the nice close button nestled right up against the border that we have in the current story book:

4px (current)
Screen Shot 2019-05-23 at 3 45 20 PM

8px
Screen Shot 2019-05-23 at 3 45 46 PM

This looks a lot better for ordinary icons like `Calendar`, but much worse for `CloseAlt`

schillerk:
Now that we're making this a breaking change, I'll go ahead and remove `icon` entirely.

Any thoughts on the before/after asymmetry? Specifically:
- Only `before` can take profileImageSrc
- Only `after` can render ButtonOrLink

This makes sense given the use cases that we've seen, but I'm happy to give both sides all functionality unless we want to intentionally restrict developers to this pattern.

stefhatcher:
> - Only `before` can take profileImageSrc
> - Only `after` can render ButtonOrLink
> 
> @schillerk 

still not sure what to do about the padding, lol. the 8px throws the x icon (probably the most common use case) off a lot. can it be 8px on the left and 4px on the right?

milesj:
I created the `v2` branch. https://github.com/airbnb/lunar/pull/67/files

Can you update this PR to merge into `v2` instead of `master`. You can do so by editing the title and then change the branch in the dropdown.

schillerk:
@stefhatcher sorry for the late reply, I realized while testing things again that even though it's called `beforeIcon`, it's really just any React Node. So you could pretty easily wrap the icon in `Spacing` if you wanted more room.

So we can go with `4px` on both sides with the understanding that users might add more spacing on their own to match the mocks, in that case, @milesj it might make sense to rename these props back to `before` and `after`?

milesj:
@schillerk Add these explicit prop types for it: https://github.com/airbnb/lunar/blob/master/packages/core/src/prop-types/iconComponent.ts

milesj:
Can you update the base branch to `v2` please.

schillerk:
@milesj got it, sorry for the delay.

Wouldn't those propTypes prevent users from wrapping the Icon with additional Spacing?

milesj:
@schillerk Consumers shouldn't be adding spacing IMO. The components are meant to be used as-is, so we should design for the most common use case.

schillerk:
Okay sounds good, I'll go with @stefhatcher 's original suggestion and set padding to `4px` for `iconAfter` to preserve the nice `CloseAlt` design, and increase padding for `iconBefore` to `8px` to better approximate the mocks.

schillerk:
Screenshots for new padding, last one should be unchanged.

Screen Shot 2019-06-11 at 10 53 22 AM
Screen Shot 2019-06-11 at 10 53 28 AM
Screen Shot 2019-06-11 at 10 53 38 AM

schillerk:
@milesj This should be ready for you to merge

milesj:
@schillerk You'll need to rebase v2 into your branch since this landed: https://github.com/airbnb/lunar/pull/86

schillerk:
@milesj sorry for the delay, will take care of this right away

schillerk:
@milesj done

packages/core/src/components/Chip/index.tsx
```diff
@@ -28,18 +30,28 @@ export type Props = {
 /** Compact component that represents a snippet of information, such as a filter. */
 export class Chip extends React.Component<Props & WithStylesProps> {
   static propTypes = {
-    icon: requiredBy('onIconClick', iconComponent),
+    // icon: requiredBy('onIconClick', iconComponent),
```
milesj: Can remove this comment.

packages/core/src/components/Chip.story.tsx
```diff
@@ -27,29 +58,24 @@ storiesOf('Core/Chip', module)
       </Spacing>
     </>
   ))
-  .add('With an icon.', () => <Chip icon={<IconCloseAlt decorative size="2em" />}>Chip</Chip>)
   .add('With an icon button.', () => (
     <>
       <Spacing right={1} inline>
-        <Chip icon={<IconCloseAlt decorative size="2em" />} onIconClick={action('onIconClick')}>
+        <Chip afterIcon={<IconCloseAlt size="2em" />} onIconClick={action('onIconClick')}>
```
milesj: Why'd you remove `decorative` from all the icons? They will now log warnings to the console.

packages/core/src/components/Chip/index.tsx
```diff
@@ -56,44 +68,53 @@ export class Chip extends React.Component<Props & WithStylesProps> {
...
-              <ProfilePhoto imageSrc={profileImageSrc} title="" size={4} />
+            <div
+              className={cx(
+                styles.sideContentInner,
+                !!beforeIcon && styles.iconWrapper,
+                !!beforeIcon && styles.iconWrapperBefore,
+              )}
+            >
+              {profileImageSrc && <ProfilePhoto imageSrc={profileImageSrc} title="" size={4} />}
+              {beforeIcon && beforeIcon}
```
milesj: Can just be `{beforeIcon}`.

packages/core/src/components/Chip/index.tsx
```diff
@@ -56,44 +68,53 @@ export class Chip extends React.Component<Props & WithStylesProps> {
...
+        {afterIcon && (
           <div className={cx(styles.chipItem, styles.sideContent)}>
-            <div className={cx(styles.sideContentInner, styles.iconWrapper)}>
+            <div
+              className={cx(styles.sideContentInner, styles.iconWrapper, styles.iconWrapperAfter)}
+            >
               {onIconClick ? (
                 <ButtonOrLink
-                  className={cx(styles.iconButton, disabled && styles.iconButton_disabled)}
+                  className={(styles.iconButton, disabled && styles.iconButton_disabled)}
```
milesj: You removed the `cx` on accident.

schillerk:
@milesj got it, thanks for all your help!

milesj:
Looks like a bad/outdated rebase again?

PR #70, opened 2019-05-24
update(DataTable): Rerender when data updates

schillerk:

## Description
There's currently a bug where DataTable does not re-render when it's passed new data. This is pretty critical since developers are often passing in data from an API call asynchronously (i.e. passing in an empty array, and then populating the table once the query resolves).

`getDerivedStateFromProps` is understandably controversial, but I think this is the proper use case.

https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html
`getDerivedStateFromProps exists for only one purpose. It enables a component to update its internal state as the result of changes in props.`

In this case, the comparison conditional is
`if (indexDataList(props.data!) != state.sortedDataList) {`
indexDataList is a little bit slow, but I've tried it with 10,000 rows and seen very little latency. Of course it depends on how complex your table is, what renderers you're using, etc.

The alternative way to do this would be with some kind of data id/version prop that the parent is responsible for updating, which is described here:
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#preferred-solutions

Of course the other issue here is that we have to run the check whenever any prop changes. I don't think we should be updating any of these props without updating the underlying data as well, but it's probably worth trying to document this somewhere more explicitly. Currently the best resource is the design doc:
https://docs.google.com/document/d/19aimBiXgzWEgECGuNPRrq_Cz4EzYIkXYlsyl_noGuS0/edit
But some kind of Readme might be helpful.

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/index.tsx
+++ b/packages/core/src/components/DataTable/index.tsx
@@ -21,7 +21,7 @@ import renderExpandableColumn from './columns/ExpandableColumn';
 import renderSelectableColumn from './columns/SelectableColumn';
 import TableHeader from './TableHeader';
 import withStyles, { css, WithStylesProps } from '../../composers/withStyles';
-import { getRowColor, getHeight } from './helpers';
+import { getRowColor, getHeight, getKeys } from './helpers';
 import { HEIGHT_TO_PX, SELECTION_OPTIONS } from './constants';
 
 export type State = {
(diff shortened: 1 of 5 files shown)


packages/core/src/components/DataTable/index.tsx
```diff
@@ -111,6 +111,14 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
     outline: 'none',
   });
 
+  static getDerivedStateFromProps(props: DataTableProps, state: State) {
+    if (indexDataList(props.data!) != state.sortedDataList) {
```
conglei: I think here the condition will be always `true`..  And we don't want to call `indexDataList(props.data)` twice (here once, the line underneath the second time).

conglei: If the performance is major consideration, we could memoize the result using for example https://github.com/caiogondim/fast-memoize.js/

schillerk: In storybook I ran a test with ~2k rows and had no issues, but it's definitely a consideration and something I'm a little worried about.

If a user/developer does start hitting this, I think the biggest improvement would be optional data version numbers:
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#preferred-solutions

Otherwise even with memoization I think we're having to compare/hash the entire data input which is faster, but still similar to the current method.

conglei: What I meant here is that `indexDataList(props.data)` will be always not equal to `state.sortedDataList`, since they are two different array.

schillerk:
Better fix here:
https://github.com/airbnb/lunar/pull/70

PR #71, opened 2019-05-29
update(Proofreader): Add background styling for important prop

al-chen:

## Description

Adding the option for a red/pink background for the proofreader composer with the `important` prop. 

## Motivation and Context

There's a need for using the proofreader component with the red/pink/`danger` background color - our designer has also approved the red-on-red underline, which gives more contrast, as seen in the screenshot below. 

## Testing

I've verified the change looks good from previewing Styleguide. I've added a unit test to ensure `important` gets plumbed through to the child `BaseTextArea`.

## Screenshots

Screen Shot 2019-05-28 at 5 15 43 PM

## Checklist


packages/core/src/components/TextArea/Proofreader/index.tsx
```diff
@@ -690,13 +695,26 @@ export class Proofreader extends React.Component<Props & WithStylesProps, State,
...
     const { unit } = theme;
 
     // Add space for controls
     const inputPadding = unit * 2; // pattern.regularButton horizontal padding
     const paddingBottom = inputPadding + unit * 4;
 
+    const { backgroundColor: colorImportant } = inputImportant;
+
+    const controlsAfter = {
+      content: '""',
```
milesj: I don't think you need to duplicate these properties, since important will just inherit from the base styles.

milesj:
For testing, at minimum you can probably test that the `important` prop is passed around.

packages/core/src/components/TextArea/Proofreader/index.tsx
```diff
@@ -727,6 +734,10 @@ export default withStyles(
         paddingBottom,
       },
 
+      highlights_important: {
```
stefhatcher: how are these styles different from `BaseTextArea`'s `important`?

packages/core/src/components/TextArea/Proofreader/index.tsx
```diff
@@ -765,6 +776,14 @@ export default withStyles(
         },
       },
 
+      controls_important: {
```
stefhatcher: aah to give controls some coloring too 👍

al-chen:
PTAL @milesj

stefhatcher:
@al-chen merged!

PR #73, opened 2019-05-30
new(DataTable): Enable Filtering and Fix Rerender Bugs

schillerk:

## Description
There was a previously documented bug with rerendering:
https://github.com/airbnb/lunar/pull/70

I wrote a solution there, but this version is much more flexible and performant, and gives us the added bonus of providing native support for filtering.

We want the table to re-render when it receives new data, but when data is being filtered through a search box or pills, we don't want to lose selection/expansion state.

There are two layers of memoization:
First, getData memoizes based on data, sortBy and sortDirection. Unless all three are unchanged, getData will produce a new indexed and sorted data list.

Within getData, the indexData function is memoized only on the data itself.

The usage of `componentDidUpdate` is not entirely ideal here, but I've read the docs and don't think there are any major problems with this aside from it being a little unconventional. This functionality could also happen inside `getDerivedStateFromProps`, but that isn't really more conventional anyway.

Because data doesn't persist in `DataTable` state anymore, it doesn't make sense to handle editing inside `DataTable` itself. Callbacks still remain, and as demonstrated in `An editable table` it is still possible to use custom renderers as editable cells.

## Testing
In storybook, the new demo `A table with a search box.` demonstrates all these capabilities.

## Checklist


schillerk:
@conglei

packages/core/src/components/DataTable.story.tsx
```diff
@@ -1,17 +1,24 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
+// @ts-ignore
```
milesj: What is this for?

packages/core/src/components/DataTable/DefaultRenderer.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
-import Text from '../../Text';
-import { RendererProps } from '../types';
+import Text from '../Text';
+import { RendererProps } from './types';
 
-export default function TextRenderer({ row, key }: RendererProps) {
+export default function renderDefaultContent({ row, key, onEdit }: RendererProps) {
```
milesj: `onEdit` not being used?

packages/core/src/components/DataTable.story.tsx
schillerk: I'm getting what I thought was just a storybook bug:

```
Could not find a declaration file for module '@airbnb/lunar-icons/lib/interface/IconStar'. '/Users/kyle_schiller/lunar/packages/icons/lib/interface/IconStar.js' implicitly has an 'any' type.
  Try `npm install @types/airbnb__lunar-icons` if it exists or add a new declaration (.d.ts) file containing `declare module '@airbnb/lunar-icons/lib/interface/IconStar';`ts(7016)
```

packages/core/src/components/DataTable/helpers/indexData.ts
```diff
@@ -0,0 +1,36 @@
+import memoize from 'lodash/memoize';
+
+import { ChildRow, IndexedChildRow, IndexedParentRow, ParentRow } from '../types';
+
+// eslint-disable-next-line import/prefer-default-export
+export const indexData = memoize(
```
milesj: Why isnt default being used?

packages/core/src/components/DataTable/index.tsx
```diff
@@ -164,57 +164,28 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
         newExpandedRows.add(newExpandedRowIndex);
       }
 
+      console.log(newExpandedRows);
```
milesj: Remove

milesj:
Seeing as how this is a breaking change, it'll have to land as part of the 2.0 release.

packages/core/src/components/DataTable/helpers/indexData.ts
schillerk: Since it's memorized it needs to be exported as a const instead of a function, which is invalid:
https://stackoverflow.com/questions/36261225/why-is-export-default-const-invalid

milesj:
I'll defer to @conglei for the logic changes.

schillerk:
I'm pretty sure it's only a breaking change for users who had `instantEdit` set to `false`. I can't be  sure, but I'm not aware of anyone doing this and didn't see it in any of the use cases we looked at. In general I'm not aware of anyone using DataTable outside of DX.

milesj:
@schillerk Doesn't matter who the consumers are, it's still a breaking API change and is treated as such.

Tgeo:
Hey @milesj the version I'm using currently `1.11.0` has a bug which causes `DataTable` to not refresh when provided with a new set of `data` rows. This causes `DataTable` to be unusable for my use-case (paginating 20 rows of data at a time, clicking through to the next page gets the next 20 results, etc).

What is the best way to get this fix out the door?

milesj:
@Tgeo Not be a breaking change. He'll have to revert all the edit removal stuff.

Tgeo:
@milesj any way you could take a look at @schillerk's changes today?

packages/core/src/components/DataTable.story.tsx
```diff
@@ -82,6 +93,60 @@ const defaultEditCallback = (
...
+  };
+
+  private handleNewData = () => {
+    this.setState({
+      data: generateRandomData(),
+    });
+  };
+
+  filter = (search: string) => (data: IndexedParentRow[]): IndexedParentRow[] => {
+    // @ts-ignore
```
stefhatcher: what's this for?

.storybook/components/DataTable/DataTableRenderers/EditableTextRenderer.jsx
```diff
@@ -0,0 +1,42 @@
...
+    });
+  };
+
+  render() {
+    const { editMode, row, key } = this.props;
+    const { value } = this.state;
+
+    return editMode ? (
+      <Input
+        label=""
```
stefhatcher: how about a label like "Edit row"? Maybe with the row number? having a label is important for accessibility, even if it is visually hidden

.storybook/components/DataTable/DataTableRenderers/EditableTextRenderer.jsx
```diff
@@ -0,0 +1,42 @@
...
+        name=""
+        hideLabel
```
stefhatcher: please move `hideLabel ` to the front to be consistent with other boolean flag prop sorting.

packages/core/src/components/DataTable.story.tsx
```diff
@@ -82,6 +93,60 @@ const defaultEditCallback = (
...
+      <Button inline onClick={this.handleNewData}>
+        New Data
+      </Button>
+    );
+
+    return (
+      <>
+        <Spacing bottom={2}>
+          <Row before={button}>
+            <Input inline label="" name="" hideLabel value={search} onChange={this.handleChange} />
```
stefhatcher: same comments about `label` and `hideLabel`

packages/core/src/components/DataTable/helpers.ts
```diff
@@ -58,3 +66,20 @@ export function getHeight(defaultHeight?: RowHeightOptions, overrideHeight?: Hei
 
   return defaultHeight ? HEIGHT_TO_PX[defaultHeight] : 0;
 }
+
+// Infers keys from data if they aren't explicitely defined
```
stefhatcher: `explicitly`

packages/core/test/components/DataTable.test.tsx
```diff
@@ -3,16 +3,80 @@ import { shallow, mount } from 'enzyme';
...
+      value: newVal,
+    });
+  };
+
+  render() {
+    const { editMode, row, keyName } = this.props;
+    const { value } = this.state;
+
+    return editMode ? (
+      <Input label="" name="" hideLabel value={value} onChange={this.onEdit(row, keyName)} />
```
stefhatcher: same comments about `label` and `hideLabel`

packages/core/src/components/DataTable.story.tsx
schillerk: There's an example use case in the `A table with a search box` story.

Previously I intended for users to write any kind of filter logic in the container component and just update the data prop each time.

In this new model, updating the underlying data implies that an entirely new data set has actually been added, which means we reset selected/expanded state. This would be undesirable if a user is just searching through an input field.

Letting filtering happen inside the data table means we can handle that kind of use case without actually updating the `data` prop, and get to have selected/expanded state persist.

schillerk:
@stefhatcher Hope the explanation on the new `filter` prop makes sense! Just pushed fixes for all the other comments

packages/core/src/components/DataTable.story.tsx
stefhatcher: aah! that really helps. i know this is just a `story`, but maybe a small comment around this ignore/what this is letting happen would help people looking at the story code? sounds like: let filtering happen without mutating `data` and persist state ?

schillerk: Got it, I'll go ahead and add that, thanks for flagging this!

schillerk:
Hey @stefhatcher, re: Mile's previous comments, I reverted previous changes so that this fixes the known existing bugs in DataTable without being a breaking change. The API is entirely intact and `instantEdit` is still a valid prop.

Would it be possible to get this approved and merged so @Tgeo can use the updated DataTable?

stefhatcher:
aah! wasn't sure about that. i'll take another look @schillerk ! and dismiss his blocking review

schillerk:
> aah! wasn't sure about that. i'll take another look @schillerk ! and dismiss his blocking review

Thanks so much!!

packages/core/src/components/DataTable/TableHeader.tsx
```diff
@@ -11,10 +11,10 @@ export type Props = {
   editable?: boolean;
   /** Determines which set of header buttons to render. */
   editMode: boolean;
-  /** Without instantEdit the header renders Cancel and Apply, otherwise it just renders Done. */
-  instantEdit?: boolean;
   /** Height of the TableHeader, falls back to RowHeight if not specified. */
   height: number;
+  /** If instantEdit is disabled, header will render Cancel and Apply buttons during edit mode. */
```
stefhatcher: this description tripped me up a little - thinking it was `true` by default. how about `Renders a Done button. Defaults to Cancel and Apply buttons.`

packages/core/test/components/DataTable.test.tsx
```diff
@@ -483,47 +575,38 @@ describe('<DataTable /> handles edits', () => {
...
+        .find(Input)
+        .prop('value'),
+    ).toBe(data[0].data.name);
   });
 
-  it('should be editable', () => {
+  it('should be editable without instant edit', () => {
     // @ts-ignore
-    const wrapper = mount(<DataTable {...props} />);
+    const wrapper = mount(<DataTable {...props} instantEdit={false} />);
```
stefhatcher: not seeing it - but is `instantEdit` true in `props`?

packages/core/src/components/DataTable/TableHeader.tsx
schillerk: Oh yeah, `instantEdit` does default to `true`, I like your description more though, I'll go ahead and update to that

packages/core/test/components/DataTable.test.tsx
schillerk: Yeah, `instantEdit` is default `true`. That happens in `DataTable/index` under `static defaultProps`

packages/core/test/components/DataTable.test.tsx
```diff
@@ -466,7 +558,7 @@ describe('<DataTable /> handles edits', () => {
   });
 
   it('should enable instant edit mode', () => {
-    const wrapper = shallow(<DataTable {...props} instantEdit />).dive();
+    const wrapper = shallow(<DataTable {...props} />).dive();
```
stefhatcher: should have `instantEdit` on it? not seeing it defined in `props` as `true`

schillerk: That was an unnecessary prop in the previous version, `instantEdit` is and always was default `true`

packages/core/src/components/DataTable/TableHeader.tsx
stefhatcher: where does it default to true? i don't see that being set in this file

packages/core/src/components/DataTable/TableHeader.tsx
```diff
@@ -23,9 +23,7 @@ export type Props = {
   onEnableEditMode: () => void;
   /** Callback for toggling editMode. */
   onDisableEditMode: () => void;
-  /** Undos edits if instantEdit is disabled. */
-  onCancelEditMode: () => void;
-  /** Applys edits if instantEdit is enabled. */
+  /** Applys edits if instantEdit is dissabled. */
```
stefhatcher: `Applies` & `disabled`

packages/core/test/components/DataTable.test.tsx
stefhatcher: aaah, k. i see it in the `DataTable/index` now used to seeing defaults where the props are defined / cause changes like that. with a default of false, usually, and true when the flag is set. defaultProps sometimes renders the falsey (undefined) value first before applying default prop values.

packages/core/test/components/DataTable.test.tsx
stefhatcher: aah, gotcha!

packages/core/src/components/DataTable/TableHeader.tsx
schillerk: got it, thanks!

schillerk:
@stefhatcher anything else I can clean up that's blocking this from landing?

stefhatcher:
@schillerk looks good! other than the update to describing `instantEdit`! but not a blocker

stefhatcher:
@schillerk i updated it with master and was able to dismiss miles's review. @milesj - your concerns looked addressed! hope that's okay

schillerk:
@stefhatcher oh weird, not sure why it didn't update inline, the description for `instantEdit` does not read `  /** Renders a Done button. Defaults to Cancel and Apply buttons. */`

I don't have write access so you'll have to merge this into master.

Thank you so much!!!

stefhatcher:
@schillerk merged!

PR #74, opened 2019-05-31
style: Migrate globals and styles from O2 

milesj:

## Description

- Adds wrapping logic/capabilities for text and tables.
- Adds more global styles that are extremely useful.
- Support locale aware fonts.

## Motivation and Context

Feature parity with our old O2 system.

## Testing

Checked against the style guide.

## Checklist

### DIFF
--- a/packages/core/src/components/DatePickerInput/Input/index.tsx
+++ b/packages/core/src/components/DatePickerInput/Input/index.tsx
@@ -50,7 +50,7 @@ export default class PrivatePickerInput extends DayPickerInput {
     this.input = ref;
 
     if (propagateRef) {
-      propagateRef(ref);
+      propagateRef.current = ref;
     }
   };
 
--- a/packages/core/src/components/Rating/index.tsx
+++ b/packages/core/src/components/Rating/index.tsx
@@ -64,7 +64,7 @@ export class Rating extends React.Component<Props & WithStylesProps> {
               Star = IconStarFull;
             }
 
-            return <Star key={key} color={color} decorative />;
+            return <Star key={key} color={color} size="1.15em" decorative />;
           })}
 
           {reviews && <Spacing left={reviews ? 0.5 : 0}>{reviews}</Spacing>}
--- a/packages/core/src/components/Table/Cell.tsx
+++ b/packages/core/src/components/Table/Cell.tsx
@@ -14,6 +14,8 @@ export type Props = {
   startAlign?: boolean;
   /** Truncate text and display all on hover. */
   truncate?: boolean;
+  /** Wrap text and white space. */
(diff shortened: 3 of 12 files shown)


packages/core/src/components/Table/Cell.tsx
```diff
@@ -14,6 +14,8 @@ export type Props = {
   startAlign?: boolean;
   /** Truncate text and display all on hover. */
   truncate?: boolean;
+  /** Wrap text and white space. */
+  wrap?: boolean;
```
stefhatcher: aaah - okay! i was curious if `wrap` was a thing (when looking at your other PR using this)

packages/core/src/themes/global.ts
```diff
@@ -1,7 +1,7 @@
 import { FontFace, GlobalSheet } from 'aesthetic';
 import { Theme } from '../types';
 
-export default (fontFaces: { [fontFamily: string]: FontFace[] }) => ({ color }: Theme) =>
+export default (fontFaces: { [fontFamily: string]: FontFace[] }) => ({ color, font }: Theme) =>
   ({
     '@global': {
       '*': {
```
stefhatcher: would this be a good time to finally add:

```css
:focus:not(:focus-visible) { outline: none }
```

Gets rid of the annoying outline for mouse users but preserves it for keyboard users, and is ignored by browsers that don’t support `:focus-visible`.

packages/core/src/components/DatePickerInput/Input/index.tsx
```diff
@@ -50,7 +50,7 @@ export default class PrivatePickerInput extends DayPickerInput {
     this.input = ref;
 
     if (propagateRef) {
-      propagateRef(ref);
+      propagateRef.current = ref;
```
milesj: Was breaking storybook.

packages/core/src/components/Rating/index.tsx
```diff
@@ -64,7 +64,7 @@ export class Rating extends React.Component<Props & WithStylesProps> {
               Star = IconStarFull;
             }
 
-            return <Star key={key} color={color} decorative />;
+            return <Star key={key} color={color} size="1.15em" decorative />;
```
stefhatcher: why this change? `Rating` already has size options that were based off of `1em`

packages/core/src/themes/global.ts
stefhatcher: i see you've added it! resolving this!

packages/core/src/components/Rating/index.tsx
milesj: Slightly scaled up to match font size.

PR #78, opened 2019-06-05
update(Pagination): Add pageLabel prop

piewell:

## Description

Add pageLabel prop so "Page" can be changed to a different text value.

## Motivation and Context

While working on the Lightbox I needed a pagination component without the word "Page". Not all pagination is strictly for pages - we may need to change the label from "Page" to "Photo" or "".
Example:
`pageLabel="Photo"` results in "Photo x" / "Photo x of x"
`pageLabel=""` results in "x" / "x of x"

## Testing

- Added to Storybook
- Tested locally

## Screenshots

Screen Shot 2019-06-05 at 2 25 00 PM
Screen Shot 2019-06-05 at 2 25 04 PM

## Checklist

### DIFF
--- a/packages/core/src/components/Pagination.story.tsx
+++ b/packages/core/src/components/Pagination.story.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
 import Pagination from './Pagination';
+import T from './Translate';
 
 storiesOf('Core/Pagination', module)
   .addParameters({
@@ -20,6 +21,16 @@ storiesOf('Core/Pagination', module)
     <Pagination
       hasPrev
       hasNext
+      page={2}
+      onNext={action('onNext')}
+      onPrevious={action('onPrevious')}
+    />
+  ))
+  .add('With label.', () => (
+    <Pagination
+      hasPrev
+      hasNext
+      label={T.phrase('Photo', {}, 'Label for photo pagination')}
       page={2}
       onNext={action('onNext')}
       onPrevious={action('onPrevious')}
--- a/packages/core/src/components/Pagination.story.tsx
+++ b/packages/core/src/components/Pagination.story.tsx
@@ -30,7 +30,17 @@ storiesOf('Core/Pagination', module)
     <Pagination
       hasPrev
       hasNext
-      label={T.phrase('Photo', {}, 'Label for photo pagination')}
+      pageLabel={T.phrase('Photo', {}, 'Label for photo pagination')}
+      page={2}
+      onNext={action('onNext')}
+      onPrevious={action('onPrevious')}
+    />
+  ))
+  .add('With no label.', () => (
+    <Pagination
+      hasPrev
+      hasNext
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Pagination.story.tsx
```diff
@@ -25,6 +26,26 @@ storiesOf('Core/Pagination', module)
...
+      page={2}
+      onNext={action('onNext')}
+      onPrevious={action('onPrevious')}
+    />
+  ))
+  .add('With no label.', () => (
+    <Pagination
+      hasPrev
+      hasNext
+      pageLabel=""
```
stefhatcher: since the type is `string | undefined` (optional), let's removed `pageLabel=""` from the example.

packages/core/src/components/Pagination/index.tsx
```diff
@@ -150,14 +154,16 @@ export class Pagination extends React.Component<Props & WithStylesProps> {
           <Text muted>
             {showBookends && pageCount ? (
               <T
-                phrase="Page %{pageNumber} of %{pageCount}"
+                phrase="%{pageLabel} %{pageNumber} of %{pageCount}"
```
stefhatcher: and since `pageLabel?:` is optional - it can have `undefined` as a value, so we need translate without the `pageLabel` variable. 

something like (pseudo code):

```js
let paginationText = showBookends && pageCount ? 
  phrase="%{pageNumber} of %{pageCount}"  : 
  page;

if (pageLabel) {
  paginationText = showBookends && pageCount ?
    phrase="%{pageLabel} %{pageNumber} of %{pageCount}" : 
    phrase="%{pageLabel} %{pageNumber}";
}
```

stefhatcher: or make pageLabel required (empty string and no undefined) `pageLabel: string;` in the prop defs and have no default prop value

packages/core/src/components/Pagination.story.tsx
piewell: So I ran into some trouble here while trying to avoid a breaking change. 😬 If we want no label, but the default is "Page", is there another way to have no label? Maybe a `noLabel` prop, or is that redundant?

stefhatcher: aaah i see! so `undefined` can get the defaultProp `Page` and when `pageLabel=""`, do the translation phrases without it as a variable

packages/core/src/components/Pagination/index.tsx
```diff
@@ -130,6 +134,38 @@ export class Pagination extends React.Component<Props & WithStylesProps> {
...
+    let paginationText =
+      showBookends && pageCount ? (
+        <T
+          phrase={'%{pageNumber} of %{pageNumber}'}
+          pageCount={pageCount}
+          pageNumber={page}
+          context="Showing the current page number and total page count"
+        />
+      ) : (
+        <T phrase={'%{pageNumber}'} pageNumber={page} context="Showing the current page number" />
```
milesj: This phrase should be remove IMO. Just use `pageNumber` directly.

PR #82, opened 2019-06-07
update(Proofreader): Add `isRuleHighlighted` and `isRuleSecondary` props. Add `SecondaryMark` component.

al-chen:
cc: @qectu @tyn2

## Description

1. Update the `ProofreadRuleMatch` type to expect a `rule_id` that the proofreader endpoint will deliver. The rule will determine what the state of highlighting and color. 
2. Add the `alwaysHighlight` prop to `Mark`. When true, all errors are always highlighted. 
3. Add the `SecondaryMark` component, which extends `Mark` and styles the marked color as yellow (`color.core.warning[2]` to be precise). 

There's one minor discrepancy in styling between `Mark` and `SecondaryMark`, other than color - in `Mark`, opacity is set to 0.75 for underlines and `ui.disabledOpacity` for highlights. In `SecondaryMark`, opacity is set to be always 0.75. This is intentional. 

See screenshots for details. 

Mark is used exclusively by Proofreader, so I'm marking this as an `update(Proofreader)`. 

## Motivation and Context

There is concern that only underlining does not bring enough attention to certain, high-priority errors that demand engagement. Highlighting from the beginning can be a different strategy that differentiates from underlining. 

The motivation for secondary coloring is similar; it should signal a different level of engagement to users who might be used to equating red to, for example, purely spellchecking or trivial errors.

## Testing

Updated and written new unit tests. Verified UI changes in local storybook. 

## Screenshots
Screen Shot 2019-06-06 at 5 10 07 PM

Screen Shot 2019-06-06 at 5 09 55 PM

Screen Shot 2019-06-06 at 5 10 46 PM

Screen Shot 2019-06-06 at 5 08 03 PM

Screen Shot 2019-06-06 at 5 07 45 PM

Screen Shot 2019-06-06 at 5 10 57 PM

## Checklist


packages/core/src/components/TextArea/index.tsx
```diff
@@ -30,9 +32,11 @@ export default class TextArea extends React.Component<Props, State> {
   };
 
   static defaultProps = {
+    alwaysHighlight: false,
```
milesj: I'm not a fan of adding these extra props to `TextArea`. Add a new `proofreaderProps` prop that is an object of props to pass down.

packages/core/src/components/TextArea/index.tsx
```diff
@@ -30,9 +32,11 @@ export default class TextArea extends React.Component<Props, State> {
...
     locale: 'none',
     noTranslate: false,
     proofread: false,
+    secondaryMark: false,
```
milesj: This seems weird. So the consumer toggles between the colors manually? It's not automated?

packages/core/src/components/TextArea/Proofreader/index.tsx
```diff
@@ -37,15 +38,17 @@ export type Position = {
...
-  locale?: string;
-  name: string;
-  id: string;
-  noTranslate?: boolean;
-  onChange: (value: string, event: React.ChangeEvent<HTMLTextAreaElement>) => void;
-  onCheckText: (params: any) => Promise<ProofreaderResponse>;
-  value: string;
-};
+export type Props = Pick<FormInputProps, 'important'> &
+  Partial<Pick<MarkProps, 'alwaysHighlight'>> & {
```
milesj: Instead of picking, you can just reference the other type.

```
alwaysHighlight?: MarkProps['alwaysHighlight'];
```

packages/core/src/components/TextArea/Proofreader/Mark.tsx
```diff
@@ -5,6 +5,7 @@ export type Props = {
   children: NonNullable<React.ReactNode>;
   selected: boolean;
   onSelect: (top: number, left: number) => void;
+  alwaysHighlight: boolean;
```
milesj: Should be optional.

packages/core/src/components/TextArea/constants.ts
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/prefer-default-export
+export const LANGUAGE_RULES = {
+  UNNECESSARY_ADMIN_NOTE_RULE: 'AGENT_TOOLS_UNNECESSARY_ADMIN_NOTE_RULE',
```
milesj: This is tying third-party code to an open source project, so will not be accepted. I think the best solution would be to add new props to proofreader that check for secondary/highlighted state, for example.

```ts
isRuleHighlighted?: (rule: ProofreadRuleMatch) => boolean;
isRuleSecondary?: (rule: ProofreadRuleMatch) => boolean;
```

This is a similar implmentation to props on `Autocomplete`.

al-chen:
@milesj Addressed, PTAL

packages/core/src/components/TextArea/index.tsx
```diff
@@ -58,7 +62,14 @@ export default class TextArea extends React.Component<Props, State> {
     return (
       <FormField {...fieldProps} id={id} labelDescription={description}>
         {proofread ? (
-          <Proofreader {...(textareaProps as any)} id={id} locale={locale} />
+          <Proofreader
+            {...(textareaProps as any)}
+            {...proofreadProps}
+            id={id}
+            locale={locale}
+            isRuleHighlighted={proofreadProps && proofreadProps.isRuleHighlighted}
```
milesj: You don't need these 2 lines if you're spreading `proofreadProps` above.

al-chen: Oh that's right, thanks!

PR #83, opened 2019-06-07
update(Proofreader): Handle delete replacement

al-chen:
cc: @QecTu @tyn2

## Description

Handle the case in Proofreader when error replacement is `''` by representing it with `(Delete)`. 

## Motivation and Context

When error replacement is an empty string, it ostensibly means we should delete the error. We should make that more obvious rather than display a blank row. 

## Testing

Updated unit tests. Verified locally in storybook. 

## Screenshots
Before:
Screen Shot 2019-06-06 at 6 04 11 PM

After:
Jun-07-2019 10-51-51

## Checklist


tyn2:
@al-chen Looks good, can we just update the text to `Delete this note`?

packages/core/src/components/TextArea/Proofreader/ErrorMenu.tsx
```diff
@@ -16,6 +16,17 @@ export default class ErrorMenu extends React.Component<Props> {
     this.props.onReplaceText(this.props.error, replacement);
   };
 
+  private replacementPhrase = (replacement: string) => {
```
stefhatcher: is this `private` needed? usually just added for handlers

al-chen: It doesn't seem necessary, but to me (admittedly neither a TS nor Lunar expert), I'm not sure there's harm in adding it to encapsulate implementation within this component. Can remove if convention and style disagree.

tyn2:
@al-chen Can we actually do `Delete this note` without the parentheses?

packages/core/src/components/TextArea/Proofreader/ErrorMenu.tsx
stefhatcher: not a blocker! did a rough scan of lunar and it is mostly in front of handlers, but some other methods too. seems fine!

milesj:
Why don't we just filter out errors that are `''`?

al-chen:
@milesj this change isn't for errors that are `''`, it's for replacements to the errors that are `''`. Seems intuitive to me that if you want to replace text with `''`, you _probably_ want to delete the original text - we're making that more obvious here.

milesj:
@al-chen The after image helps. Curious where the "No need to note this action..." message come from? Not sure "note" makes enough sense in this context.

tyn2:
@milesj @al-chen The `No need to note this action…` tooltip is from our content strategist and I'm assuming "note" is taken derived from "admin notes". Or do you mean the `Delete this note` text is ambiguous?

al-chen:
@milesj @stefhatcher hello! if this LG, can we merge?

Issue #26, opened 2019-04-10
Autocomplete component issues


mmontag:
1. It seems that `loadItemsOnMount` implies `autoFocus`. This leads to some undesirable behavior if, for example, you apply this to multiple Autocomplete components on the same page.

2. Autocomplete component does not fire `onLoadOptions` callback for empty string input. (The only time `onLoadOptions` is fired for an empty string is when `loadItemsOnMount` is `true`.)

3. The built-in caching 5 minutes of autocomplete items cannot be disabled or configured.

The combination of issue 2 and 3 leads to some other edge cases.

Finally, just a side remark, the naming seems a bit inconsistent, as in `loadItemsOnMount` vs. `onLoadOptions`. Could be more clear if this was **items** everywhere.

mmontag:
cc @amisha-patel

milesj:
@mmontag 

1) This is intended. It would be weird to display the autocomplete as "open" if you're not focused on it. Native selects don't work that way.

2) This is a bug, will fix.

3) Will add a prop to bypass cache.

mmontag:
Ok, thanks. I was confused about the purpose of `loadItemsOnMount`. Thought it was just concerned with preloading items.

williaster:
@milesj we are also facing the issue described by `#1`. We want to pre-fetch the `Autocomplete` options for an empty query so that they display _immediately_ upon focus, which I feel like is a valid use-case

Can I fix this / add an option to opt-out of this behavior?

williaster:
Actually I think we aren't even that concerned about the pre-fetching, we would be happy if we just had a way to fetch items _on focus_ (to populate a set of popular options) instead of requiring the user to enter something

mmontag:
Echoing @williaster— our current solution is to tell users to type a single space to force the list of all options to show up.

milesj:
@mmontag 

This is an alternative for 1: https://github.com/airbnb/lunar/pull/85
This should fix 3: https://github.com/airbnb/lunar/pull/91

The 2nd item won't be fixed as it's quite a behavior change in the component. Can use `loadItemsOnFocus` for possible empty string logic.

mmontag:
Thanks @milesj, I'll try this out soon.

milesj:
V2 landed which included some fixes. Gonna close this for now. If there are other issues, feel free to open another issue.

PR #85, opened 2019-06-10
new(Autocomplete): Add `loadItemsOnFocus` functionality

williaster:

## Description

This PR adds a new `loadItemsOnFocus` prop to the `Autocomplete` (and thus `Multicomplete`) component(s). 

## Motivation and Context

The new `loadItemsOnFocus` partially addresses some component limitations described in #26 ([comment](https://github.com/airbnb/lunar/issues/26#issuecomment-496802807)), namely that it is impossible to fetch items upon which we've designed several user experiences around. 

We have tried using the `autoFocus` and `loadItemsOnMount` props, but they do not enable the same functionality and user experience.

## Testing

Functional (`@airbnb/lunar-core` `Autocomplete`, `Multicomplete`, and `@airbnb/lunar-form` `Autocomplete`) + new unit test.

## Screenshots

**Before** (must type to load)

**After** (load on focus)

Multicomplete

## Checklist

### DIFF
--- a/packages/core/src/components/Autocomplete.story.tsx
+++ b/packages/core/src/components/Autocomplete.story.tsx
@@ -67,11 +67,11 @@ storiesOf('Core/Autocomplete', module)
       accessibilityLabel="Label"
       name="autocomplete-load-on-focus"
       label="Label"
-      loadItemsOnFocus
       onChange={action('onChange')}
       onSelectItem={action('onSelectItem')}
       onLoadOptions={value => Promise.resolve(items) as any}
       labelDescription="Load some items on focus."
+      loadItemsOnFocus
     />
   ))
   .add('Disable selected items with `isItemSelectable`.', () => (
--- a/packages/core/src/components/Autocomplete.story.tsx
+++ b/packages/core/src/components/Autocomplete.story.tsx
@@ -62,6 +62,18 @@ storiesOf('Core/Autocomplete', module)
       disabled
     />
   ))
+  .add('Load items on focus.', () => (
+    <Autocomplete
+      accessibilityLabel="Label"
+      name="autocomplete-load-on-focus"
+      label="Label"
+      loadItemsOnFocus
+      onChange={action('onChange')}
+      onSelectItem={action('onSelectItem')}
+      onLoadOptions={value => Promise.resolve(items) as any}
+      labelDescription="Load some items on focus."
+    />
+  ))
   .add('Disable selected items with `isItemSelectable`.', () => (
     <Autocomplete
       accessibilityLabel="Favorite color?"
--- a/packages/core/src/components/Autocomplete/index.tsx
+++ b/packages/core/src/components/Autocomplete/index.tsx
@@ -50,6 +50,8 @@ export type Props<T extends Item> = Omit<BaseInputProps, 'id'> &
     isItemSelectable?: (item: T, selected?: boolean) => boolean;
     /** Determine if an item is selected. Will compare values by default if not defined. */
     isItemSelected?: (item: T, value: string) => boolean;
+    /** */
+    loadItemsOnFocus?: boolean;
     /** Load and show items on mount. */
     loadItemsOnMount?: boolean;
     /** Max height of the results dropdown menu. */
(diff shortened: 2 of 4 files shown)


packages/core/src/components/Autocomplete/index.tsx
```diff
@@ -600,7 +609,7 @@ export default class Autocomplete<T extends Item> extends React.Component<Props<
     const { error, loading, value } = this.state;
     const items = this.renderItems();
 
-    if (!value && items.length === 0) {
+    if (!loading && !value && items.length === 0) {
```
williaster: previously this line would hide the `menu` if there was no input `value`

packages/core/src/components/Autocomplete.story.tsx
```diff
@@ -62,6 +62,18 @@ storiesOf('Core/Autocomplete', module)
       disabled
     />
   ))
+  .add('Load items on focus.', () => (
+    <Autocomplete
+      accessibilityLabel="Label"
+      name="autocomplete-load-on-focus"
+      label="Label"
+      loadItemsOnFocus
```
stefhatcher: please move `loadItemsOnFocus` boolean prop to the bottom to be consistent with the rest of the story

packages/core/test/components/Autocomplete.test.tsx
```diff
@@ -164,6 +164,18 @@ describe('<Autocomplete />', () => {
 
       expect(wrapper.state('open')).toBe(true);
     });
+
+    it('calls loadItems if `loadItemsOnFocus` is `true`', () => {
```
stefhatcher: mind adding this test to Multicomplete too?

packages/core/src/components/Multicomplete.story.tsx
```diff
@@ -48,4 +48,26 @@ storiesOf('Core/Multicomplete', module)
...
+            { value: 'red', name: 'Red' },
+            { value: 'black', name: 'Black' },
+            { value: 'blue', name: 'Blue' },
+            { value: 'green', name: 'Green' },
+          ].filter(item => !value || item.name.toLowerCase().match(value.toLowerCase())),
+        )
+      }
+      renderItem={(item, highlighted, selected) => <Text bold={selected}>{item.name}</Text>}
+      value={['red', 'green']}
+      loadItemsOnFocus
```
stefhatcher: can you add a screenshot of this working for Multicomplete too?

packages/core/test/components/Autocomplete.test.tsx
williaster: doing this is pretty gross because 
- `loadItems` is not an instance method of `Multicomplete` so we can't spy on it
- we can't get a handle on the `autocomplete` instance because it's not the parent of `shallow()`
- spying on `onLoadOptions` requires a `timeout` because it's debounced 😭

packages/core/src/components/Autocomplete/index.tsx
```diff
@@ -219,6 +222,10 @@ export default class Autocomplete<T extends Item> extends React.Component<Props<
       this.props.onFocus(event);
     }
 
+    if (this.props.loadItemsOnFocus) {
+      this.loadItems('');
```
milesj: Maybe this should be the current value and not an empty string?

packages/core/test/components/Autocomplete.test.tsx
williaster: Currently we don't verify all of the other `Autocomplete` functionality in the `Multicomplete` tests, I feel like this isn't really necessary since the functionality lives in `Autocomplete` and we have tests there.

stefhatcher: oof - fair! wasn't sure and haven't checked this branch out locally. trying to cover all the bases that Multicomplete is 👍 too

packages/core/src/components/Multicomplete.story.tsx
williaster: done (above)

packages/core/src/components/Autocomplete/index.tsx
williaster: good call when there's a value already else results would be wrong 👍

williaster:
@stefhatcher @milesj comments addressed / rebased on latest.

PR #87, opened 2019-06-11
new(SidePanel): Add new collapsible side panel.

schillerk:

## Description

This adds a new SidePanel component to Layouts, most of the code is pulled directly from Bighead, but we would like to reuse it across DX projects.

There's some overlap in functionality between this and the TwoColumnLayout. The two main additions are:
- Collapsibility
- Percent width constrained by min/max values

There are props to modify the size, color, top offset and compactness of the collapse button.

It would be more flexible to let developers pass in an arbitrary node as icon, but we're designing for the most common use cases, and this allows us to use the more stringent `iconComponent` type check.

## Screenshots
Screen Shot 2019-06-11 at 11 05 00 AM

Screen Shot 2019-06-11 at 12 01 23 PM

Collapsed:
Screen Shot 2019-06-11 at 12 05 24 PM

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "name": "lunar-root",
-  "version": "0.1.0",
+  "version": "2.0.0",
   "description": "React toolkit and design language for Airbnb open source and internal projects.",
   "license": "MIT",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -85,10 +85,16 @@
     "@babel/preset-env": "^7.4.5",
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.3.3",
+<<<<<<< HEAD
     "@types/enzyme": "^3.9.4",
     "@types/google.analytics": "0.0.40",
     "@types/jest": "^24.0.15",
     "@types/jscodeshift": "^0.6.1",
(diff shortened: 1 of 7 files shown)


milesj:
Before I review this, is this meant to be stand alone? Or actually used within the layout components? As it currently stands, the aside width is controlled by the layout.

schillerk:
Hey sorry, yeah, this is meant to be used on its own. It doesn't actually interact with `Layout`

milesj:
I'm going to make a breaking change in layouts for v2. Let's land that and then re-visit this as it may change how this API works.

schillerk:
Oh wait sorry, why does the change to layouts matter if this doesn't interact with `Layout`?

milesj:
Because the new `Aside` component achieves half of what this does. They could possibly be merged into 1. Progress here: https://github.com/airbnb/lunar/pull/95

schillerk:
Ah cool, yeah sounds good!

stefhatcher:
@schillerk @milesj can this be closed in favor of https://github.com/airbnb/lunar/pull/229 ?

stefhatcher:
closing in favor of https://github.com/airbnb/lunar/pull/229

PR #88, opened 2019-06-11
new(AdaptiveGrid): Add new AdaptiveGrid

schillerk:

## Description
AdaptiveGrid wraps items and lays them out in a grid of equal width columns and variable height rows. Items overflow at the end of rows, and flow from left to right.

The `breakpoints` prop allows developers to define behavior at different screen widths. For example, with `breakpoints={1200: 5, 1000: 3}`, the grid will display 5 items per row for window widths greater than 1200 pixels, 3 items per row for window widths greater than 1000 pixels, and `defaultItems` items per row otherwise.

The `padding` prop allows developers to define padding in units in between grid items. without affecting padding around the grid.

## Screenshots
Screen Shot 2019-06-11 at 1 17 27 PM

## Checklist


packages/core/src/components/DynamicGrid/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+  padding?: number;
+};
+
+type Props = GridProps & WithStylesProps;
+
+class DynamicGrid extends React.PureComponent<Props, State> {
+  public static defaultProps = {
+    breakpoints: {},
+    defaultItems: 1,
+    padding: 2,
```
stefhatcher: hrmm maybe no `padding` prop so the padding between columns can be made consistent with `Grid`'s `Col` padding? a `noPadding` prop if no padding is desired

schillerk: You could set `padding` to `0`, does that work?
Screen Shot 2019-06-11 at 1 47 35 PM

stefhatcher: i'm suggesting no `padding` prop and baking in the same column padding as Grid/Col components. `noPadding` was another suggestion if that's a use case you would need for this. my concern is with keeping grid spacing consistent within the design system.

schillerk:
@stefhatcher sorry for the delay, what would you think about renaming the prop `customPadding` or `paddingOverride` and changing the description to something like `// Forces override of Lunar's consistent 2 unit padding`?

If that still seems too flexible I'm fine with the `noPadding` boolean option.

stefhatcher:
yeah, i'd like the consistent padding with a `noPadding` option please! layout components have the same pattern coming in v2

milesj:
Not sure the naming is the best. Instead of padding we should use gutter, and the component name should maybe be something like `AdaptiveGrid` or `UniformGrid`?

packages/core/src/components/DynamicGrid/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+type State = {
+  items: number;
+};
+
+export type GridProps = {
+  /** List of width/item pairs, describes how many items to display for windows of at least that width. */
+  breakpoints?: { [key: number]: number };
+  /** Items per row for screens smaller than the smallest breakpoint. */
+  defaultItems?: number;
```
milesj: `itemCount`?

packages/core/src/components/DynamicGrid/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+type Props = GridProps & WithStylesProps;
+
+class DynamicGrid extends React.PureComponent<Props, State> {
+  public static defaultProps = {
+    breakpoints: {},
+    defaultItems: 1,
+    noPadding: false,
+  };
+
+  constructor(props: Props) {
```
milesj: Dont need the constructor.

```
state = {
  items: this.getItems(),
};
```

packages/core/src/components/DynamicGrid/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+  /** Items per row for screens smaller than the smallest breakpoint. */
+  defaultItems?: number;
+  /** Removes padding between items. */
+  noPadding?: boolean;
+};
+
+type Props = GridProps & WithStylesProps;
+
+class DynamicGrid extends React.PureComponent<Props, State> {
+  public static defaultProps = {
```
milesj: Remove the visibility modifiers from everything.

packages/core/src/components/DynamicGrid/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+  componentDidMount() {
+    window.addEventListener('resize', this.handleResize);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.handleResize);
+  }
+
+  // Find the largest breakpoint smaller than the current width
+  private getItems(): any {
```
milesj: Dont use `any`.

packages/core/src/components/DynamicGrid/index.tsx
schillerk: `defaultItemsPerRow` or `defaultItemCount`?

schillerk:
Thanks for the feedback, went with `AdaptiveGrid`

packages/core/src/components/DynamicGrid/index.tsx
milesj: `defaultItemCount` works 👍

packages/core/src/components/AdaptiveGrid/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const { children, cx, noGutter, styles } = this.props;
+    const { items } = this.state;
+
+    const itemStyle = {
+      width: `${100 / items}%`,
+    };
+
+    const childElements =
+      children &&
+      React.Children.map(children, (child: React.ReactNode, idx: number) => (
```
milesj: We should check that child is truthy, otherwise we'd be wrapping an empty value with a `div`, which is probably not what we want.

```
if (!child) return null;
```

schillerk:
@milesj @stefhatcher I think this should be ready now

packages/core/src/components/AdaptiveGrid/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+    defaultItemsPerRow: 1,
+    noGutter: false,
+  };
+
+  state = {
+    items: this.getItems(),
+  };
+
+  componentDidMount() {
+    window.addEventListener('resize', this.handleResize);
```
stefhatcher: i wish we could solve this without listening for the resize event :/ can you explore taking the breakpoints and setting the styles based on that? css grid might help here - per media query, setting the css grid column count or something. `cx` can handle media queries, right @milesj ?

milesj: Yeah it can. Supports any CSS.

schillerk:
@stefhatcher I updated this to use CSS grid, but am not sure how to do breakpoints based on props without a resize listener. I can add media queries to the `withStyles` block, but don't have access to component props in `withStyles` right?

There is sort of a way to do this with CSS grid `autosize` and `minmax`, but as I understand it those only work in the case where you have maximum 3 different row sizes, and the smallest is always 1 item. To be fair, that does cover a lot of use cases.

If it helps at all, the current component is a `PureComponent`, so at least it's not rerendering on resize unless items per row actually needs to update.

schillerk:
Incidentally this makes the tests fail because we can no longer pull width off children, but I'll leave that until I figure out what implementation is actually going to look like.

stefhatcher:
i'm curious if something like this would work (doing the media queries in cx and my example has the breakpoint data structured a little differently for demo):

```jsx
// in the example, passing in the `breakpoints`
breakpoints={[{ width: 1200, items: 5 }, { width: 800, items: 1 }]}

// render() {
const { breakpoints } = this.props; 
let media = {};

// build up the media queries
// max is 0px -> max. might need to sort the width from largest to smallest first ?
breakpoints.forEach(breakpoint => media[`(max-width: ${breakpoint.width}px)`] = { gridTemplateColumns: `repeat(${breakpoint.items}, 1fr)` });

// in return {
// inline after display grid styles
className={styles.grid, { '@media':  { styles.media } } }
```
that's might be it ^ haven't tested it out

packages/core/src/components/AdaptiveGrid/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+    breakpointStyles[
+      breakpointKeys.length
+        ? `@media (max-width: ${smallestBreakpoint}px)`
+        : `@media (min-width: 0px)`
+    ] = {
+      gridTemplateColumns: `repeat(${defaultItemsPerRow}, 1fr)`,
+    };
+
+    console.log('rendering');
```
stefhatcher: dont forget to remove this

packages/core/src/components/AdaptiveGrid/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    return (
+      <div className={cx(styles.container, !noGutter && styles.container_padded, breakpointStyles)}>
+        {childElements}
+      </div>
+    );
+  }
+}
+export default withStyles(({ unit, responsive }) => ({
+  container: {
+    display: 'grid',
```
stefhatcher: can add `gridGap: unit *2` here and do away with `container_padded` and `item_padded`

then, when `noGutter` is true:
```
container_noGutter: {
  gridGap: 0,
}
```

packages/core/src/components/AdaptiveGrid/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+        ? `@media (max-width: ${smallestBreakpoint}px)`
+        : `@media (min-width: 0px)`
+    ] = {
+      gridTemplateColumns: `repeat(${defaultItemsPerRow}, 1fr)`,
+    };
+
+    console.log('rendering');
+
+    return (
+      <div className={cx(styles.container, !noGutter && styles.container_padded, breakpointStyles)}>
```
stefhatcher: with my suggestion below, this would be `cx(styles.container, noGutter && styles.container_noGutter breakpointStyles)` instead

stefhatcher:
yay! the css solution worked out!!!

schillerk:
@stefhatcher would you mind making the merge? Thanks!

stefhatcher:
@schillerk i would like the gridGap update too please? that way css grid handles all the gutter and no negative margins are needed

stefhatcher:
shoot - your base branch needs to change to (point to master) now that v2 has been merged.

schillerk:
Woo got it, thanks for that, also sorry, just realized you're on PTO 😭

packages/core/src/components/AdaptiveGrid/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+    breakpointKeys.forEach(breakpoint => {
+      breakpointStyles[`@media (min-width: ${breakpoint}px)`] = {
+        gridTemplateColumns: `repeat(${breakpoints![breakpoint]}, 1fr)`,
+      };
+    });
+
+    breakpointStyles[
+      breakpointKeys.length > 0
+        ? `@media (max-width: ${smallestBreakpoint}px)`
+        : `@media (min-width: 0px)`
```
stefhatcher: nit: doesn't need string template here `'@media (min-width: 0px)'`

packages/core/src/components/AdaptiveGrid.story.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import AdaptiveGrid from './AdaptiveGrid';
+
+const gridItem = (
+  <div style={{ height: '100%', border: '1px solid black', padding: '16px' }}>Grid Item</div>
+);
+const bigGridItem = (
+  <div style={{ height: '100%', border: '1px solid black', padding: '16px' }}>
+    Very Big Grid Item Very Big Grid Item Very Big Grid Item Very Big Grid Item Very Big Grid Item
```
milesj: Use `<LoremIpsum />` component here?

stefhatcher: it has a `short` prop too for smaller ipsum

packages/core/src/components/AdaptiveGrid/index.tsx
schillerk: Hm this didn't update inline, but I just updated to make this change;
https://github.com/airbnb/lunar/pull/88/files

Thanks!

stefhatcher:
thanks @schillerk ! i'll merge after the builds are done and pass

PR #92, opened 2019-06-12
break(Tabs): Rework active state. Add small, beforeIcon, and afterIcon props.

milesj:

## Description

Adds an active state that colors text to our primary blurple. This technically isn't a breaking change (except the theme changes below) as the API is the same, but the active state will be broken in consumers based on what label they are passing to `Tab`.

- [x] Add support for icons.
- [x] Add support for small text.
- [x] Add an active state.

### Migration Guide

- If you were relying on `Text` for font sizing and color, you should remove them and simply pass a string. Otherwise the active state color will not be used.

```ts
// Before
<Tab key="key" label={<Text>Label</Text>} />
<Tab key="key" label={<Text small>Label</Text>} />

// After
<Tab key="key" label="Label" />
<Tab key="key" label="Label" small />
```

- The theme `borderFocus` accent property has been renamed.

```ts
// Before
theme.color.accent.borderFocus;

// After
theme.color.accent.borderActive;
```

## Motivation and Context

We've been wanting a true active state for a while now, so this is it.

## Testing

Unit tests and storybook.

## Screenshots

Screen Shot 2019-06-12 at 12 44 52 PM
Screen Shot 2019-06-12 at 12 58 11 PM
Screen Shot 2019-06-12 at 12 58 15 PM

## Checklist

### DIFF
--- a/packages/core/src/components/Tabs/Tab.tsx
+++ b/packages/core/src/components/Tabs/Tab.tsx
@@ -23,6 +23,8 @@ export type Props = Pick<ButtonOrLinkProps, 'afterIcon' | 'beforeIcon' | 'disabl
   onSelected?: () => void;
   /** Whether the tab is selected or not. */
   selected?: boolean;
+  /** Decrease font size to small. */
+  small?: boolean;
   /** Stretch tabs to fill the full width. */
   stretched?: boolean;
 };
@@ -33,6 +35,7 @@ export class Tab extends React.Component<Props & WithStylesProps> {
     borderless: false,
     children: null,
     selected: false,
+    small: false,
     stretched: false,
   };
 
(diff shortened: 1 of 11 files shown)


packages/core/src/components/Tabs/Tab.tsx
```diff
@@ -137,9 +144,12 @@ export default withStyles(({ color, font, pattern, unit, ui, transition }) => ({
 
   tabButton: {
     ...font.textReset,
+    ...font.textRegular,
     ...transition.box,
     background: color.accent.bg,
-    display: 'inline-block',
+    color: color.accent.text,
+    display: 'flex',
+    alignItems: 'center',
```
stefhatcher: `justifyContent: 'center'` too?

PR #95, opened 2019-06-13
break(Layouts): Rework layout structuring. Add LayoutShell and Aside components.

milesj:

## Description

The current layout architecture is very basic and not at all useful for large applications. It was designed in such a way that the `*Layout` component must always be rendered at the root of the application, since it includes the side/top bars. This is quite problematic once apps become to complex, rely on crazy router logic, and need to adjust their layout based on many factors, but persist the sidebar.

This rework accomplishes the following:

- The side/top bar navigation logic has moved to a new component, `LayoutShell`, which should be rendered near the root. Now these bars will always persist across pages, regardless of how the content/router changes. 

- Asides should now be wrapped in a new `Aside` component, which handles the width, padding, and styling of the side columns. This opens up many future possibilities, for example, collapsible asides, scrollable content, variable width columns, and much more.

### Migration Guide

- Aside columns must be wrapped with `Aside`. The `asideWidth` prop has moved from all layout components to this new component.

```ts
// Before
<Layout asideWidth={300} before={<Column />} after={<Column />} />

// After
<Layout 
  before={<Aside width={300}><Column /></Aside>} 
  after={<Aside width={300}><Column /></Aside>} 
/>
```

- The `sideBar` prop has moved from all layout components to the new `LayoutShell` component.

```ts
// Before
<Layout sideBar={<SideBar />} />

// After
<LayoutShell sideBar={<SideBar />} />
  <Layout />
</LayoutShell>
```

## Motivation and Context

We want more control of how layouts are structured.

## Testing

Storybook.

## Screenshots

Screen Shot 2019-06-14 at 1 26 48 PM
Screen Shot 2019-06-14 at 1 26 58 PM
Screen Shot 2019-06-14 at 1 27 11 PM

## Checklist

### DIFF
--- a/packages/layouts/src/components/LayoutShell/index.tsx
+++ b/packages/layouts/src/components/LayoutShell/index.tsx
@@ -28,11 +28,12 @@ export type Props = {
   // topBar?: React.ReactNode;
 };
 
+/** Layout shell that wraps an entire application, providing optional side and top nav bars. */
 export default function LayoutShell({ children, sideBar }: Props) {
   const [styles, cx] = useStyles(styleSheet);
 
   return (
-    <div className={cx(styles.layout)}>
+    <div className={cx(styles.shell)}>
       {sideBar && <aside className={cx(styles.sideBar)}>{sideBar}</aside>}
 
       <div>{children}</div>
--- a/packages/layouts/src/components/LayoutShell/index.tsx
+++ b/packages/layouts/src/components/LayoutShell/index.tsx
@@ -1,0 +1,45 @@
+import React from 'react';
+import { elementType } from 'airbnb-prop-types';
+import useStyles, { Theme } from '@airbnb/lunar/lib/hooks/useStyles';
+import SideBar from '../SideBar';
+
+const styleSheet = ({ color }: Theme) => ({
+  shell: {
+    display: 'flex',
+    width: '100%',
+    minHeight: '100vh',
+    background: color.accent.bg,
+    justifyContent: 'space-between',
+  },
+
+  sideBar: {
+    flexGrow: 0,
+    flexShrink: 0,
+    width: 'auto',
+  },
+});
+
+export type Props = {
+  /** The page content. */
+  children: NonNullable<React.ReactNode>;
+  /** Navigation side bar to display before the content. */
+  sideBar?: React.ReactNode;
+  /** Navigation top bar to display above the content. */
+  // topBar?: React.ReactNode;
+};
+
+export default function LayoutShell({ children, sideBar }: Props) {
+  const [styles, cx] = useStyles(styleSheet);
+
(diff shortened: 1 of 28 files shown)


packages/layouts/src/components/LayoutShell/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  children: NonNullable<React.ReactNode>;
+  /** Navigation side bar to display before the content. */
+  sideBar?: React.ReactNode;
+  /** Navigation top bar to display above the content. */
+  // topBar?: React.ReactNode;
+};
+
+/** Layout shell that wraps an entire application, providing optional side and top nav bars. */
+function LayoutShell({ children, sideBar }: Props) {
+  const [styles, cx] = useStyles(styleSheet);
```
stefhatcher: can we start using stylesheet hooks for new stuff? or is this for packages outside of core?

packages/layouts/test/components/Layout.test.tsx
```diff
@@ -18,44 +12,44 @@ describe('<Layout />', () => {
...
+  it('renders with props passed', () => {
+    const wrapper = shallow(
+      <Layout noBackground noPadding>
+        Child
+      </Layout>,
+    ).dive();
 
-    expect(wrapper.find('aside')).toHaveLength(1);
-    expect(wrapper.find('aside').contains(sidebar)).toBe(true);
+    expect(wrapper.find('main')).toHaveLength(1);
```
stefhatcher: doesn't look like this test is testing the props as passed. update the description?

packages/layouts/src/components/LayoutShell/index.tsx
milesj: Yeah we can use hooks.

packages/layouts/src/components/SplitLayout/index.tsx
```diff
@@ -1,41 +1,39 @@
...
   wrapper: {
     display: 'flex',
     height: '100%',
   },
 
   column: {
     width: '50%',
-    flexGrow: 0,
-    flexShrink: 0,
-    padding: unit * 2,
```
stefhatcher: add column padding back and `noPadding` prop? that'd keep layout gutters consistent

stefhatcher: or did the padding move into the before/after components? checking

stefhatcher: it did! resolving myself

packages/layouts/test/components/Layout.test.tsx
milesj: Technically it still is testing the additional props for code coverage, but we can't run expects on styles/class names.

PR #114, opened 2019-07-10
update(AccordionItem): Change title to node instead of string

davidchang:

## Description

AccordionItem expects a `title` of string, but this just changes the prop type to more broadly accept a ReactNode.

## Motivation and Context

I had a use case in which I wanted to use AccordionItem, but specifically needed to have a more complex title.

## Testing

```npm run jest```

and visual check of Storybook story.

## Screenshots

Storybook now, with a story and updated `title` prop in the table (showing ReactNode instead of string)

image

## Checklist

### DIFF
--- a/packages/core/src/components/Accordion/Item.tsx
+++ b/packages/core/src/components/Accordion/Item.tsx
@@ -14,7 +14,7 @@ export type Props = {
   /** Index amongst a collection of accordion items. */
   index?: number;
   /** Title of the accordion item. */
-  title?: string;
+  title?: React.ReactNode;
   /** Callback fired when the accordion item is clicked. */
   onClick?: (index: number) => void;
 };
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/core/src/components/Accordion/Item.tsx [hunk 1]
```diff
@@ -14,7 +14,7 @@ export type Props = {
   /** Index amongst a collection of accordion items. */
   index?: number;
   /** Title of the accordion item. */
-  title?: string;
+  title?: React.ReactNode;
```
stefhatcher: maybe update the type to be `string | React.ReactNode`

davidchang: i originally had that, but i noticed Storybook only showed `ReactNode` as the type. then i thought that `string` is actually encapsulated under `ReactNode`? https://basarat.gitbooks.io/typescript/docs/jsx/react.html seems to verify (if you search ReactNode)

stefhatcher: dope! forgot about that 👍

davidchang:
^ if this is okay, @stefhatcher , could you merge? i don't have write access

stefhatcher:
thanks @davidchang ! merged

PR #117, opened 2019-07-11
test: Integrate `TestAesthetic` for easier class name testing

milesj:

## Description

Made some changes to Aesthetic upstream to make testing easier: https://milesj.gitbook.io/aesthetic/testing 

This allows us to use deterministic class names and to stop using snapshots.

## Motivation and Context

We don't like using snapshots.

## Testing

Of course!

## Screenshots

N/A

## Checklist


packages/core/src/components/private/BaseSelect.tsx
```diff
@@ -36,7 +36,7 @@ class BaseSelect extends React.Component<Props & WithStylesProps> {
           {children}
         </FormInput>
 
-        <div
+        <i
```
stefhatcher: why `i`? instead of `span` if you want it to be inline?

packages/core/test/components/Accordion.test.tsx
```diff
@@ -43,7 +43,7 @@ describe('<Accordion />', () => {
       id: 0,
     });
 
-    expect(wrapper).toMatchSnapshot();
+    expect(wrapper.prop('className')).toMatch('container_bordered');
```
stefhatcher: this is such a nice way to replace snapshots 🙏

packages/core/src/components/private/BaseSelect.tsx
milesj: `i` = icons.

stefhatcher: aah and i guess it wouldn't matter if the svg is italic

(i = italic)

milesj: Honestly, I mainly did it so testing was easier hah, but I have noticed icon libraries use this pattern.

PR #113, opened 2019-07-09
style(Button,Input,Autocomplete,Menu): Normalize and standardize spacing

schillerk:
to: @milesj @stefhatcher @hayes @alecklandgraf

## Description
Changes:
- Decrease `large` spacing from `20px` to `16px`
- Add `large` option to `Input`
- Increase `small` spacing from `6/10` to `8/12`
- Change `MenuRow` in `Autocomplete` to `MenuItem` for No Results and Error states
- Add Spacing to `Autocomplete` Loading state
- Add top/bottom padding to `Menu`
- Subtract border width from spacing calculations
- **[new as of 7/11]** Increase theme borderRadius to 4

A few controversial points:
- The `Input` size props are now `compact` and `large`, ideally this would be `small` and large. I could add the `small` prop with redundant functionality and add a developers note that `compact` will be deprecated and/or we could make this a breaking change in the future.

- First/Last Menu Items now have asymmetrical focus outline. This is to make the extra padding clickable as part of the items.
Screen Shot 2019-07-09 at 11 46 48 AM

- When `Button` is borderless, spacing doesn't grow to compensate

## Motivation and Context
@kenchendesign has noticed some inconsistencies and misalignments in Lunar spacing these changes aim to correct. There's a full description of these issues in another PR I'll try to find and append.

## Screenshots
Screen Shot 2019-07-09 at 1 51 54 PM
Screen Shot 2019-07-09 at 1 51 49 PM
Screen Shot 2019-07-09 at 1 51 38 PM
Screen Shot 2019-07-09 at 1 51 18 PM
Screen Shot 2019-07-09 at 1 51 08 PM
Screen Shot 2019-07-09 at 1 50 58 PM

## Checklist

- My code follows the style guide of this project.
- I have updated or added documentation accordingly.
- I have read the CONTRIBUTING document.

PR #118, opened 2019-07-11
style(Button): Adjust minimum widths.

schillerk:

## Description
This reduces min-width on each of the button sizes and removes the golden ratio calculation.

Values were previously:
52, 78, 104
Screen Shot 2019-07-11 at 11 40 04 AM

Values are now:
48, 64, 72
Screen Shot 2019-07-11 at 11 39 55 AM

## Motivation and Context
See discussion between @adamirl and @kenchendesign in #dls-for-internaltools

## Checklist

### DIFF
--- a/packages/core/src/components/Button/index.tsx
+++ b/packages/core/src/components/Button/index.tsx
@@ -6,8 +6,6 @@ import Loader from '../Loader';
 
 const sizingProp = mutuallyExclusiveTrueProps('small', 'large');
 
-const GOLDEN_RATIO = (1 + Math.sqrt(5)) / 2;
-
 export type Props = ButtonOrLinkProps & {
   /** Render as a block with full width. */
   block?: boolean;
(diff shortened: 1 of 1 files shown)


schillerk:
Going to wait for https://github.com/airbnb/lunar/pull/117 to land before updating tests.
Will probably be rewriting Button tests for
https://github.com/airbnb/lunar/pull/113#pullrequestreview-260867874 anyway

milesj:
I kind of like the golden ratio. Why not just lower the unit value used with the ratio?

schillerk:
@milesj Yeah, I would have liked to preserve it too, but then we end up with
48 / (1.618 * 8) = 3.7
64 / (1.618 * 8) = 4.94
72 / (1.618 * 8) = 5.56

So we could make 64 work (5 * 1.6 * 8), but

`small` would have to be ~51 (4 * 1.6 * 8) or 38 (3 * 1.6 * 8) and

`large` would have to be ~77 (6 * 1.6 * 8)

@kenchendesign ?

kenchendesign:
I don’t think the golden ratio is meaningful anymore if you need to add
another multiplier.

milesj:
What is the importance of 48, 64, and 72?

kenchendesign:
I guess they are just nice numbers that are multiples of 8. And they
buttons look visually balanced in our experiments.

Yeah I agree it’s all subjective and I’m sure 60 would look good if 64
does. The bottom line is, since the rest of the UIs are based on multiples
of 4 so we should do 64 rather than 62. Beyond that it’s just visual
preference. We already got agreement from multiple designers on the
internal channel so I would recommend those numbers.

PR #120, opened 2019-07-12
new: Use `React.memo` under the hood. Remove children/node prop type checks.

milesj:

## Description

Noticed some performance degradation so updated Aesthetic upstream to use `React.memo` in `withStyles`. This caused some slight React tree changes that conflicted with our prop types, so I just removed our prop types as they are quite annoying.

## Motivation and Context

For the gainz.

## Checklist

### DIFF
--- a/packages/apollo/test/components/Mutation.test.tsx
+++ b/packages/apollo/test/components/Mutation.test.tsx
@@ -47,19 +47,19 @@ describe('Mutation', () => {
       },
     };
 
-    it('renders a `Loader` by default', () => {
-      const wrapper = renderer.create(
-        <MockedProvider mocks={[mock]} addTypename={false}>
-          <WrappingComponent>
-            <Mutation mutation={MUTATION}>{childHandler}</Mutation>
-          </WrappingComponent>
-        </MockedProvider>,
-      );
+    // it('renders a `Loader` by default', () => {
+    //   const wrapper = renderer.create(
+    //     <MockedProvider mocks={[mock]} addTypename={false}>
+    //       <WrappingComponent>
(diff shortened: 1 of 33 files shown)


packages/apollo/test/components/Mutation.test.tsx
```diff
@@ -47,19 +47,19 @@ describe('Mutation', () => {
...
-            <Mutation mutation={MUTATION}>{childHandler}</Mutation>
-          </WrappingComponent>
-        </MockedProvider>,
-      );
-
-      wrapper.root.findByType('button').props.onClick();
-
-      expect(wrapper.root.findByType(Loader)).toBeDefined();
-    });
+    // it('renders a `Loader` by default', () => {
```
milesj: Will fix in a follow up now that Enzyme supports context.

Alphy11: **SOON**<sup>tm</sup>

Alphy11:
@milesj does this still respect the pureComponent rule?

Alphy11:
Also, should this be a breaking change?  It could potentially break things.

milesj:
@Alphy11 The pure option in Aesthetic? That doesn't exist anymore, so everything is pure (now memo) by default. We tested the memo in Nova just now and haven't noticed any weird side effects.

Alphy11:
Also: `so I just removed our prop types as they are quite annoying.`

Alphy11:
@milesj That can cause problems if any consumers have components with side effects (e.g. if they have not upgraded react router to V5

milesj:
@Alphy11 It was pure by default before, and this is the same thing. It could potentially be a problem if pure was disabled, but I'd suggest moving to `useStyles` over `withStyles`.

PR #125, opened 2019-07-17
new: Migrate from `raven-js` to `@sentry/browser`

milesj:

## Description

Raven has been deprecated in favor of Sentry. https://github.com/getsentry/sentry-javascript

Adds new `sentry` setting for full/more control.

## Motivation and Context

Stay up to date.

## Checklist

### DIFF
--- a/packages/metrics/src/index.ts
+++ b/packages/metrics/src/index.ts
@@ -31,13 +31,13 @@ class Metrics {
 
     this.bootstrapNewRelic();
     this.bootstrapSentry();
-    this.bootstrapGoogleAnalyticsUser();
+    this.bootstrapGoogleAnalytics();
   }
 
   bootstrapNewRelic() {
     const { context, ignoreErrors, userID } = this.settings;
 
-    if (!hasNewRelic()) {
+    if (!this.isNewRelicEnabled()) {
       return;
     }
 
@@ -67,7 +67,7 @@ class Metrics {
     const { context, ignoreErrors, sentry, sentryKey, sentryProject, userID } = this.settings;
     const { host, protocol } = global.location;
 
-    if (!(sentry && sentry.dsn) && !(sentryKey && sentryProject)) {
+    if (!this.isSentryEnabled()) {
       return;
     }
 
@@ -87,10 +87,28 @@ class Metrics {
     });
   }
 
-  bootstrapGoogleAnalyticsUser() {
-    if (hasGoogleAnalytics() && this.settings.userID) {
+  bootstrapGoogleAnalytics() {
+    if (!this.isGoogleAnalyticsEnabled()) {
+      return;
+    }
+
+    if (this.settings.userID) {
       ga('set', 'userId', `${this.settings.userID}`);
     }
+  }
+
+  isGoogleAnalyticsEnabled() {
+    return hasGoogleAnalytics();
+  }
+
+  isNewRelicEnabled() {
+    return hasNewRelic();
+  }
+
+  isSentryEnabled() {
+    const { sentry, sentryKey, sentryProject } = this.settings;
+
+    return (sentry && sentry.dsn) || (sentryKey && sentryProject);
   }
 }
 
--- a/packages/metrics/src/index.ts
+++ b/packages/metrics/src/index.ts
@@ -7,6 +7,7 @@ export type IgnoreError = string | RegExp;
 export type Settings = {
   context?: { [key: string]: unknown };
   ignoreErrors?: IgnoreError[];
+  sentryDSN?: string;
   sentryKey?: string;
   sentryProject?: string;
   userID?: number | null;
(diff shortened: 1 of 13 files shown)


packages/metrics/src/index.ts
```diff
@@ -62,27 +66,38 @@ class Metrics {
...
+    if (!sentryDSN && (!sentryKey || !sentryProject)) {
       return;
     }
 
-    const ravenHost = `${global.location.protocol}//${sentryKey}@${global.location.host}`;
-    const ravenPath = `/proxy/sentry/${sentryProject}`;
-
-    Raven.config(`${ravenHost}${ravenPath}`, {
+    initSentry({
+      dsn: sentryDSN || `${protocol}//${sentryKey}@${host}/${sentryProject}`,
```
Alphy11: dsn accepts a breakdown that can make this a bit easier.
https://git.musta.ch/airbnb/torch/blob/master/src/setup/setupMetrics.js#L28

packages/metrics/src/index.ts
```diff
@@ -7,18 +7,22 @@ export type IgnoreError = string | RegExp;
 export type Settings = {
   context?: { [key: string]: unknown };
   ignoreErrors?: IgnoreError[];
+  sentryDSN?: string;
   sentryKey?: string;
   sentryProject?: string;
   userID?: number | null;
+  onSentryScope?: ((scope: Scope) => void) | null;
```
Alphy11: not super clear what the intention of this is.

packages/metrics/src/index.ts
```diff
@@ -62,27 +66,38 @@ class Metrics {
...
-      .setUserContext({
-        browserLocale: global.navigator.language,
-        userAgent: global.navigator.userAgent,
-        userID: userID || 'N/A',
-        ...context,
-      })
-      .install();
+    });
+
+    configureScope(scope => {
```
Alphy11: might be an extra PR, but we update the sentry breadcrumbs with the tracking names as well, to make them more useful.
https://git.musta.ch/airbnb/torch/blob/master/src/setup/setupMetrics.js#L20

packages/metrics/src/utils/captureError.ts
```diff
@@ -1,10 +1,7 @@
-import Raven from 'raven-js';
+import { captureException } from '@sentry/browser';
 import hasNewRelic from './hasNewRelic';
 
-export default function captureError(
-  error?: string | Event | Error | null,
-  options: Raven.RavenOptions = {},
-) {
+export default function captureError(error?: string | Event | Error | null) {
```
Alphy11: Instead of making this breaking, we could take in extras still, and automatically add them in the scope.
https://git.musta.ch/airbnb/torch/blob/master/src/shared/metrics/torchLogger.ts#L45

packages/metrics/src/index.ts
milesj: Allows consumers to hook into sentry since it's abstracted away.

```
Metrics.initialize({
  onSentryScope(scope => {
    scope.addTag('custom', 'consumer specific');
  }),
});
```

Alphy11: Any reason this is better than keeping the current interface for the logger and just converting extras to scope for sentry?  Taht way we can reuse the same things for new relic as well.

milesj: True, good point.

packages/metrics/src/utils/captureError.ts
milesj: Good idea, added `withScope`.

packages/metrics/src/utils/captureError.ts
```diff
@@ -23,5 +28,15 @@ export default function captureError(
...
 
-  Raven.captureException(errorInstance, options);
+  withScope(scope => {
+    if (options.context) {
+      scope.setContext(errorInstance.name, options.context);
+    }
+
+    if (options.extra) {
+      scope.setExtras(options.extra);
+    }
```
Alphy11: Can you add fingerprint here as well?

milesj: I'll add all the scope things.

packages/metrics/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import Raven from 'raven-js';
+import { init as initSentry, configureScope, BrowserOptions } from '@sentry/browser';
```
Alphy11: Can we export all of sentry here?  Otherwise apps can't really take advantage of most of the setup if they want to use any additional options.  For example we have an on click event that logs breadcrumbs.  There are other context events that sentry exposes as well.

milesj: I can just make `@sentry/browser` a peer dep. There's a lot of stuff in this package, so pick and choosing what to re-export seems overkill.

Alphy11: 👍 I was thinking `export * as Sentry from '@sentry/browser`, but peer dep works too.

milesj: Actually, exporting it under a namespace isn't a bad idea either.

PR #127, opened 2019-07-18
update(DataTable): add prop for callback on row selection

schillerk:

## Description
Currently, the only want to interact with selectedRows from outside DataTable is to configure custom header buttons. This satisfied the existing use case for Bighead of selecting rows and then performing an action on them, but there have been requests for having access to selections as a callback.

This PR adds a new prop for users to pass in a function, and then adds it as a `setState` callback when we update `selectedRows`

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable.story.tsx
+++ b/packages/core/src/components/DataTable.story.tsx
@@ -16,7 +16,7 @@ import Button from './Button';
 import Input from './Input';
 import Row from './Row';
 import Spacing from './Spacing';
-import { SelectedRows, VirtualRow, IndexedParentRow } from './DataTable/types';
+import { ExpandedRow, SelectedRows, IndexedParentRow, VirtualRow } from './DataTable/types';
 
 const renderers = {
   name: EditableTextRenderer,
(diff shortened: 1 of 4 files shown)


packages/core/src/components/DataTable.story.tsx
```diff
@@ -234,12 +238,15 @@ storiesOf('Core/DataTable', module)
       tableHeaderHeight="large"
     />
   ))
-  .add('An table that logs custom edit callbacks.', () => (
+  .add('An table that logs custom edit callbacks and select callback.', () => (
```
stefhatcher: `An` -> `A`

packages/core/src/components/DataTable/index.tsx
```diff
@@ -55,6 +55,7 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
     renderers: {},
     rowHeight: 'regular',
     selectable: false,
+    selectCallback: (rowData: ExpandedRow, selectedRows: SelectedRows) => () => {},
```
stefhatcher: making this required is a breaking change? make it optional?

stefhatcher: nvm - it was hard to tel line the diff that it's in defaultProps

schillerk:
@stefhatcher any chance we could get this and https://github.com/airbnb/lunar/pull/128 merged?

schillerk:
@stefhatcher got it

PR #128, opened 2019-07-19
update(DataTable): Add `selectedRowsFirst` prop.

schillerk:

## Description
Independent of sorting, this allows developers to pass in a prop so that selected rows are always displayed first. Does not apply to children, but does sort selected -> hasChildrenSelected -> unselected.

## Motivation and Context
Xuan Lu requested this change, it seems reasonable and could be useful in other applications as well. The user interaction is a little janky because you select a row, and then it immediately jumps to the top.

## Screenshots
Screen Shot 2019-07-18 at 5 49 29 PM

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/helpers/expandData.ts
+++ b/packages/core/src/components/DataTable/helpers/expandData.ts
@@ -1,11 +1,13 @@
 import { SortDirectionType } from 'react-virtualized';
-import sortList from './sortData';
-import { IndexedParentRow, IndexedChildRow, ExpandedRow } from '../types';
+import sortData from './sortData';
+import { SelectedRows, IndexedParentRow, IndexedChildRow, ExpandedRow } from '../types';
 
 /*  Iterate over the sortedDataList in state to flatten out children stashed in metadata */
 export default function expandData(
   sortedData: IndexedParentRow[],
   expandedRows: Set<number>,
+  selectedRows: SelectedRows,
+  selectedRowsFirst: boolean,
   sortBy: string,
   keys: string[],
   sortDirection?: SortDirectionType,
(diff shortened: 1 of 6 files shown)


packages/core/src/components/DataTable/helpers/expandData.ts
```diff
@@ -1,11 +1,13 @@
 import { SortDirectionType } from 'react-virtualized';
-import sortList from './sortData';
-import { IndexedParentRow, IndexedChildRow, ExpandedRow } from '../types';
+import sortData from './sortData';
+import { SelectedRows, IndexedParentRow, IndexedChildRow, ExpandedRow } from '../types';
 
 /*  Iterate over the sortedDataList in state to flatten out children stashed in metadata */
 export default function expandData(
   sortedData: IndexedParentRow[],
```
milesj: Not sure for this PR, but this function has far too many arguments. Maybe it should be a single object argument with multiple properties.

packages/core/src/components/DataTable/helpers/sortData.ts
```diff
@@ -17,22 +17,22 @@ function sortAsc(a: any, b: any) {
   return 1;
 }
 
-export default function sortData(
-  list: IndexedParentRow[],
+export default function sortData<T extends GenericRow>(
```
milesj: Nice generics!

packages/core/src/components/DataTable/helpers/sortData.ts
```diff
@@ -33,3 +33,29 @@ export default function sortData<T extends GenericRow>(
...
+  list: T[],
+  keys: string[],
+  selectedRows: SelectedRows,
+  selectedRowsFirst: boolean,
+  sortBy?: string,
+  sortDirection?: SortDirectionType,
+): T[] {
+  if (selectedRowsFirst) {
+    const selectedList = list.filter((row: T) =>
```
milesj: Instead of doing 2 filter loops, I would suggest just doing a `forEach` and pushing the item to either the selected or non-selected lists.

schillerk: Good point, got it!

PR #129, opened 2019-07-19
update(Translate): Add key support. Make params and context optional.

milesj:

## Description

- Adds unique key support through the `k` prop or `key` option.
- Updates params to support undefined and null.
- Updates all our usage to use keys.

## Motivation and Context

We're moving towards a direction in which phrases will require keys, so this is the first step.

## Testing

Locally.

## Screenshots

N/A

## Checklist


alecklandgraf:
I would lean against using `key` as it's special in React.

milesj:
@alecklandgraf The prop is `k`. We only use `key` when using the function form.

packages/core/src/components/Autocomplete/index.tsx
```diff
@@ -437,7 +437,15 @@ export default class Autocomplete<T extends Item> extends React.Component<Props<
...
       optional,
       placeholder:
-        placeholder || T.phrase('Search', {}, 'Search field within an autocomplete input field'),
+        placeholder ||
+        T.phrase(
+          'Search',
+          {},
+          {
+            context: 'Search field within an autocomplete input field',
+            key: 'lunar.common.search',
```
alecklandgraf: I see, only using `key` for `T.phrase`

packages/core/src/components/EmojiPicker/index.tsx
```diff
@@ -103,36 +103,40 @@ export class EmojiPicker extends React.Component<Props & WithStylesProps> {
...
-      skinNone: T.phrase('No skin tone', {}, 'Emoji picker skin tone label'),
-      skinLight: T.phrase('Light skin tone', {}, 'Emoji picker skin tone label'),
-      skinMediumLight: T.phrase('Medium-light skin tone', {}, 'Emoji picker skin tone label'),
-      skinMedium: T.phrase('Medium skin tone', {}, 'Emoji picker skin tone label'),
-      skinMediumDark: T.phrase('Medium-dark skin tone', {}, 'Emoji picker skin tone label'),
-      skinDark: T.phrase('Dark skin tone', {}, 'Emoji picker skin tone label'),
-      search: T.phrase('Search emojis', {}, 'Emoji picker search input field label'),
-      searchA11y: T.phrase(
-        'Search for emojis by keyword',
+      recentlyUsed: T.phrase('Recently Used', {}, { key: 'lunar.emoji.recentlyUsed' }),
```
alecklandgraf: can you keep the context here too? `{ key: 'lunar.emoji.recentlyUsed', context: 'Emoji picker commonly used feature type'}`

milesj: I removed the context from the `EmojiPicker` as I feel like the phrase + key name is enough to infer it has to do with emojis.

PR #131, opened 2019-07-22
update(Divider): Add short prop

schillerk:

## Description
A lot of @kenchendesign's designs involve a short divider, it's a pretty small component and I'm happy to just write it for individual apps, but seems like it could be a good fit for this Component.
https://pages.git.musta.ch/ken-chen/airbnb-prototypes/prototype-kit/

I could alternatively make this a `width` prop that takes in an arbitrary value, but in the interest of being opinionated and consistent I think a single option is more appropriate.].

## Screenshots
It's just this line haha.
Screen Shot 2019-07-22 at 2 24 11 PM

## Checklist

### DIFF
--- a/packages/core/src/components/Divider/index.tsx
+++ b/packages/core/src/components/Divider/index.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
-import Spacing, { SpacingRange } from '../Spacing';
 import withStyles, { WithStylesProps } from '../../composers/withStyles';
+import Spacing, { SpacingRange } from '../Spacing';
 
 export type Props = {
   /** Spacing on the bottom. */
@@ -15,8 +15,8 @@ export type Props = {
 export class Divider extends React.Component<Props & WithStylesProps> {
   static defaultProps: Props = {
     bottom: 2,
-    top: 2,
     short: false,
+    top: 2,
   };
 
   render() {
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Divider/index.tsx
```diff
@@ -7,28 +7,34 @@ export type Props = {
...
   };
 
   render() {
-    const { cx, styles, bottom, top } = this.props;
+    const { cx, styles, bottom, short, top } = this.props;
 
     return (
       <Spacing bottom={bottom} top={top}>
-        <div className={cx(styles.divider)} />
+        <div className={cx(styles.divider, short && styles.divider_short)} />
```
stefhatcher: can you add a test to test that the className has changed when short is passed in ? an example would be, with `short` passed on the wrapper component:

```
expect(wrapper.prop('className')).toMatch('divider_short');
```

stefhatcher:
> It's just this line haha.

🤣 love it

packages/core/test/components/Divider.test.tsx
```diff
@@ -15,4 +15,10 @@ describe('<Divider />', () => {
     expect(wrapper.prop('bottom')).toBe(4);
     expect(wrapper.prop('top')).toBe(2);
   });
+
+  it('can pass top and bottom to spacing', () => {
```
stefhatcher: hrmm maybe a different test message here? `it('renders a short divider'`?

schillerk: Ack, meant to change this, thanks

PR #132, opened 2019-07-22
update: Refine and use explicit value types. Allow null and undefined default values.

milesj:

## Description

This PR does a few things:

- Updates `connectToForm` to accept a generic, which is the type of the wrapped component's `value` prop. This allows us to guarantee the correct types and values are being passed around.
- Made `initialValue` required and `parse` preferred options so that they can be used in conjunction with the generic mentioned previously.
- Updates `defaultValue` to accept null and undefined values, so that we can do `defaultValue={obj.name}` instead of `defaultValue={obj.name || ''}`.

## Motivation and Context

We lose a lot of type information when connecting to the form, so this tries to bridge that gap. I've been thinking about a hooks API and how it may make this process easier.

## Testing

Unit tests.

## Screenshots

N/A

## Checklist

### DIFF
--- a/packages/forms/src/components/Form/Autocomplete.tsx
+++ b/packages/forms/src/components/Form/Autocomplete.tsx
@@ -1,10 +1,14 @@
 import React from 'react';
-import BaseAutocomplete, { Props, Item } from '@airbnb/lunar/lib/components/Autocomplete';
+import BaseAutocomplete, { Props } from '@airbnb/lunar/lib/components/Autocomplete';
 import connectToForm, { ConnectToFormProps } from '../../composers/connectToForm';
+import { toString } from '../../helpers';
 
 /** `Autocomplete` automatically connected to the parent `Form`. */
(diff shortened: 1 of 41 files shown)


packages/forms/src/components/Form/Autocomplete.tsx
```diff
@@ -1,10 +1,14 @@
 import React from 'react';
-import BaseAutocomplete, { Props, Item } from '@airbnb/lunar/lib/components/Autocomplete';
+import BaseAutocomplete, { Props } from '@airbnb/lunar/lib/components/Autocomplete';
 import connectToForm, { ConnectToFormProps } from '../../composers/connectToForm';
+import { toString } from '../../helpers';
 
 /** `Autocomplete` automatically connected to the parent `Form`. */
-export function FormAutocomplete<T extends Item>(props: Props<T> & ConnectToFormProps) {
-  return <BaseAutocomplete<T> {...props} />;
```
milesj: `T` is lost when using HOCs, so removing this.

PR #133, opened 2019-07-23
fix(Tooltip): Fix cannot read property getBoundingClientRect of null.

stefhatcher:

## Description

Fixes a JS error where body content is not fully loaded (null) and `getBoundingClientRect` is called. Not sure how, but seeing this error.

## Motivation and Context

Fixing errors.

## Testing

Works still in the styleguide.

## Checklist

### DIFF
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -7,7 +7,7 @@ import NotchedBox, { NOTCH_SIZE, NOTCH_SPACING } from '../NotchedBox';
 import Text from '../Text';
 import withStyles, { WithStylesProps } from '../../composers/withStyles';
 
-const defaultTargetRect: ClientRect = {
+const EMPTY_TARGET_RECT: ClientRect = {
   bottom: 0,
   height: 0,
   left: 0,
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Tooltip/index.tsx
```diff
@@ -58,7 +67,7 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
     labelID: uuid(),
     open: false,
     tooltipHeight: 0,
-    targetRect: document.body.getBoundingClientRect(),
+    targetRect: document.body ? document.body.getBoundingClientRect() : defaultTargetRect,
```
milesj: I wonder if moving this to cDM would solve the problem.

stefhatcher: moved it!

stefhatcher:
updated @milesj

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -86,6 +86,13 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
     return null;
   }
 
+  componentDidMount() {
+    // eslint-disable-next-line
```
milesj: Add the name of the rule this is disabling.

stefhatcher: updated!

stefhatcher:
updated @milesj

PR #135, opened 2019-07-23
new(TermList,Term): Add TermList component.

schillerk:

## Description
We're using some kind of key/value display in a few places. @kenchendesign and I would love to standardize this pattern for greater consistency.

This is intended as a fairly opinionated wrapper intended to enforce a particular design pattern. There are a couple open questions about just how opinionated to make it:
- If you have multiple Links, now you're wrapping them in extra `Spacing` and allowing arbitrary input.
- In the storybook example, I use `Spacing bottom={2}` to space out vertically stacked DataItems. This could be enforced with a `DataItemGroup` wrapper, as we do with `ButtonGroup`
- Same thing with horizontally stacked items, though @kenchendesign uses 6 units of spacing in this case.
- I added the `inline` prop for horizontally stacked items, but if you're adding `Spacing`, you can just do this with the `inline` prop in `Spacing`, except the last item doesn't need actual `Spacing`. So you either get
```
          <Spacing inline right={6}>
            <DataItem label="Pending tickets">8</DataItem>
          </Spacing>
          <Spacing inline>
            <DataItem label="Resolved tickets">20</DataItem>
          </Spacing>
```
Where the last `Spacing` has no actual spacing, or 
```
          <Spacing inline right={6}>
            <DataItem label="Pending tickets">8</DataItem>
          </Spacing>
          <DataItem inline label="Resolved tickets">20</DataItem>
```
Where we have an `inline` prop that only gets used once. 

## Screenshots
Screen Shot 2019-07-23 at 3 12 35 PM
Screen Shot 2019-07-23 at 3 12 38 PM
Screen Shot 2019-07-23 at 3 12 42 PM

## Checklist


packages/core/src/components/DataItem/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  link,
+  floatLinkRight,
+  inline,
+  children,
+  cx,
+  styles,
+}: {
+  label: string;
+  link?: NonNullable<React.ReactNode>;
+  floatLinkRight?: boolean;
```
stefhatcher: instead of `right` - please use `end`. maybe just `endAlign` to be consistent with naming?

looking at the comps, i would do something like:

```jsx
<div className={cx(inline && styles.inline)}>
  <div className={cx(styles.wrapper, rightAlign && styles.wrapper_rightAlign)}>
    <Spacing right={1}>
      <Text small muted>
        {label}
      </Text>
    </Spacing>

    {link && (
      <Text small muted>
        {link}
      </Text>
    )}
  </div>

  <Text>{children}</Text>
</div>


export default withStyles(() => ({
  inline: {
    display: 'inline-block',
  },

  wrapper: {
    display: 'flex',
  },

  wrapper_rightAlign: {
    justifyContent: 'space-between',
  },
}))(DataItem);
```

this would avoid some funky stuff that sometimes happens with floats and not clearing them

milesj:
Has this gone through the official design review?

packages/core/src/components/DataItem/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  inline?: boolean;
+  children: NonNullable<React.ReactNode>;
+} & WithStylesProps) {
+  return (
+    <div className={cx(inline && styles.data_item_inline)}>
```
stefhatcher: generally the `_` is used as a separator between the base style and the modifier (so, for example `styles.link, floatRight && styles.link_floatRight`). stuff after that is camelCase again. since there is no base style data item and just an inline modifier, i'd just name the style `inline` and do `inline && styles.inline`

packages/core/src/components/DataItem/index.tsx
schillerk: Good catch, thanks!

packages/core/src/components/DataItem/index.tsx
schillerk: Oh got it, made the change.

I've been adding in base class names so that when you go to look at Elements, you can see `dataItem_inline` instead of just `inline`, but I guess that's usually not a concern?

schillerk:
@milesj This has not gone through review yet, happy to let this sit, just wanted to get the PR in

packages/core/test/components/DataItem.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+describe('<DataItem />', () => {
+  it('renders inline', () => {
+    const wrapper = shallowWithStyles(
+      <DataItem inline label="key">
+        Data
+      </DataItem>,
+    );
+
+    expect(wrapper.prop('className')).toMatch('data_item_inline');
```
stefhatcher: think you'll need to update this to `inline` instead of `data_item_inline`

stefhatcher:
revisiting this - here are my thoughts on implementation (copied from our solar issue around the data item)

hrmm what about a description list, with terms and descriptions? (`<dl>` with `<dt>` & `<dd>`)? `dd`'s children would be whatever (link, button, text). that would be the semantic approach.

- `DescriptionList` would have a `horizontal` prop like `List` for a horizontal option.
- `Term` (key) accepts `<Text>` - can be muted or bold or regular or small, whatever consumers want to use
- `Definition` (value) accepts a react node - can be anything
- `Definition` would have an `after` prop support to handle the additional action links
- all `after` content would be flushed to the right to be consistent with the rest of lunar (no "not right aligned" option - just use inline links)
- in the css, `dd + dt` would have a standard amount of spacing, like `unit * 2` to be consistent with grid gutters and most gutters

```jsx
<DescriptionList>
  <Term>Key</Term> 
  <Definition after={<Link>Airflow</Link>}>
    Value
  </Definition>

  <Term>Key</Term>
  <Definition>
    Value <Link inline>Link 1</Link> <Link inline>Link 2</Link>
  </Definition>
  <Definition>Value</Definition>
</DescriptionList>
```

```
// styles
dl: {
  'dd + dt': {
    marginTop: unit * 2,
  },
},

dt: {
  // any dt reset styles
},

dd: {
  // any dd reset styles
},
```

Ideally we'd standardize the `dt` and `dd` styles to be the same.

as a component, it would work similar to `List, { Item }`. 
e.g. `DescriptionList, { Definition, Term }`

schillerk:
I started working through a couple versions of implementation. I'm a little concerned about the developer experience for a 3 Component system. Since the primary goal is encouraging consistency, I worry that just writing
```

**Comment by stefhatcher** [comment]
ah fair comment about the developer experience! i'm concerned about the semantics under the hood, and felt that it would translate a little better being a dl/dt/dd: what the component was expecting and producing, and when it would appropriate.

i like your suggestion of putting the `value` as a children of the Term component. that works! to enforce consistency, i would say have the label be a string. so something like (super bare-bones):

```jsx
// TermList
<dl>
  {children}
</dl>

// Term
<dt>
  <Row after={after}>
    <Text small>{label}</Text>
  </Row>
</dt>

<dd>
  {children}
</dd>
```

then my css comment and example above could still apply:
> in the css, `dd + dt` would have a standard amount of spacing, like `unit * 2` to be consistent with grid gutters and most gutters

packages/core/src/components/TermList/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+  cx,
+  styles,
+}: {
+  link?: NonNullable<React.ReactNode>;
+  endAlign?: boolean;
+  inline?: boolean;
+  keyTextRegular?: boolean;
+  children: NonNullable<React.ReactNode>;
+} & WithStylesProps) {
+  return <dl className={cx(inline && styles.inline, styles.termList)}>{children}</dl>;
```
stefhatcher: aah, so this would be `styles.termList, inline && styles.termList_inline` to follow our modifier pattern more

packages/core/src/components/TermList/Term.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Text>{children}</Text>
+      </dd>
+    </div>
+  );
+}
+
+export default withStyles(() => ({
+  container: {
+    '@selectors': {
+      '> dd': {
```
stefhatcher: you can do the reset on the `dd` itself

packages/core/src/components/TermList/Term.tsx
```diff
@@ -0,0 +1,54 @@
...
+  label: string;
+  after?: NonNullable<React.ReactNode>;
+  endAlign?: boolean;
+  inline?: boolean;
+  keyTextRegular?: boolean;
+  children: NonNullable<React.ReactNode>;
+} & WithStylesProps) {
+  return (
+    <div className={cx(styles.container)}>
+      <Row inline={!endAlign} after={endAlign && after}>
```
stefhatcher: i would move this all into the `dt`, so the `div` isn't necessary. in a dl - it's expected children are dt/dd only

```jsx
<>
  <dt>
    <Row inline={!endAlign} after={after}>
      <Text small={small}>{label}</Text>
    </Row>
  </dt>

  <dd className={cx(styles.dd)}>
    <Text>{children}</Text>
  </dd>
</>
```

and some cleanup around endAlign/inline/after:
- if endAlign is true, `after` goes to the end by default, so don't  need to check
- Row's `inline` makes its children inline/close, but not the entire thing (its still block level), so don't need to to the extra `!endAlign && <Spacing>` after the dt.

packages/core/src/components/TermList/Term.tsx
```diff
@@ -0,0 +1,54 @@
...
+  label,
+  after,
+  endAlign,
+  keyTextRegular,
+  children,
+  cx,
+  styles,
+}: {
+  label: string;
+  after?: NonNullable<React.ReactNode>;
```
stefhatcher: `NonNullable` and an optional prop (`?`) are conflicting. let's make it optional and accepts are ReactNode.

`after?: React.ReactNode;`

packages/core/src/components/TermList/Term.tsx
```diff
@@ -0,0 +1,54 @@
...
+  endAlign?: boolean;
+  inline?: boolean;
+  keyTextRegular?: boolean;
```
stefhatcher: let's make regular the default and small the optional? that'd be more inline with our sizing prop patterns

packages/core/src/components/TermList/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+export default withStyles(({ unit }) => ({
+  inline: {
+    display: 'inline-block',
+  },
+  termList: {
+    // Reset dl styles
+    marginBlockStart: 0,
+    marginBlockEnd: 0,
+    '@selectors': {
+      '> div:not(:last-child) > dd': {
```
stefhatcher: with the removal of the div in my other suggestion and with the goal to add spacing between `dt`s and `dd`s, this can be: 

```jsx
'> dd + dt': {
  marginTop: unit * 2,
},
```

or (wild guess)?

```jsx
'> dt:not(:first-child)': {
  marginTop: unit * 2,
},
```

schillerk:
Noting for myself:
When inline TermList overflows, Ken requests that there by 16px of vertical spacing between items as there would be if they weren't inlined.
Screen Shot 2019-10-03 at 10 18 45 AM

packages/core/src/components/TermList/Term.tsx
schillerk: Sorry I don't totally get this, you mean something like
```
dd: {
    marginInlineStart: 0,
  },
```
?

I haven't been able to get that to work

packages/core/src/components/TermList/Term.tsx
schillerk: Not 100% sure I followed the last point, but here's what I ended up with
```
 <>
      <dt>
        <Row after={endAlign && after}>
          <Text inline small={!keyTextRegular}>
            {label}
          </Text>
          {!endAlign && (
            <Spacing inline left={1}>
              {after!}
            </Spacing>
          )}
        </Row>
      </dt>
      <dd className={cx(styles.value)}>
        <Text>{children}</Text>
      </dd>
    </>
```

packages/core/src/components/TermList/index.tsx
schillerk: Went with `'> dd:not(:last-child)'` and it seems to be working

packages/core/src/components/TermList/Term.tsx
schillerk: Agreed that `keyTextRegular` is a super weird prop name, I'm changing it to `small` with the appropriate functionality, but setting `small = true` by default since that's how DataItem is used in pretty much every use case I"m aware of

packages/core/src/components/TermList/Term.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <Text>{children}</Text>
+      </dd>
+    </>
+  );
+}
+
+export default withStyles(() => ({
+  value: {
+    // reset styles
+    marginInlineStart: 0,
```
stefhatcher: you can do this reset in index.tsx - which would help keep all the dt/dd styling in one place

packages/core/src/components/TermList/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+    // Reset dl styles
+    marginBlockStart: 0,
+    marginBlockEnd: 0,
+    '@selectors': {
+      '> dd': {
+        // Reset dd styles
+        marginInlineStart: 0,
+        '@selectors': {
+          ':not(:last-child)': {
+            // marginBottom: unit * 2,
```
stefhatcher: still need this `@selectors'` block?

stefhatcher: would be nice to have this spacing between items

packages/core/src/components/TermList/Term.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import Row from '../Row';
+import Spacing from '../Spacing';
+import Text from '../Text';
+
+export default function Term({
+  label,
+  after = '',
+  endAlign,
+  small = true,
```
stefhatcher: fudge, one last thing. usually props like this are false by default. can you update it to just `small,` - let it be falsely by default?

stefhatcher: or have it be small by default, but rename this to be `regular`, and when true, the text small below is false

`Text inline small={!regular}`

packages/core/src/components/TermList/Term.tsx
```diff
@@ -0,0 +1,40 @@
...
+  children,
+}: {
+  label: string;
+  after?: React.ReactNode;
+  endAlign?: boolean;
+  small?: boolean;
+  keyTextRegular?: boolean;
```
stefhatcher: this prop looks unused

schillerk: Got it, removed

packages/core/src/components/TermList/Term.tsx
schillerk: I think Term label text is almost always small, doing default falsely was the rationale behind callingi this prop `keyTextRegular` originally, but regardless of implementation I think the default behavior should be small text

stefhatcher: sounds good! so maybe instead of a small=true, a regular=true to override the default of small

schillerk: got it!

packages/core/src/components/TermList/Term.tsx
```diff
@@ -0,0 +1,39 @@
...
+  after = '',
+  endAlign,
+  regular,
+  children,
+}: {
+  label: string;
+  after?: React.ReactNode;
+  endAlign?: boolean;
+  regular?: boolean;
+  children: NonNullable<React.ReactNode>;
```
stefhatcher: one thing that both of these new components are missing is the props being declared above, not inline with the params and the descriptions for each prop are missing, List for example:

```jsx
export type Props = {
  /** List items. */
  children: NonNullable<React.ReactNode>;
  /** Apply gutters between `li`s. */
  gutter?: boolean;
  /** Horizontal list. */
  horizontal?: boolean;
  /** Align contents in the middle vertically, to be used with `horizontal`. */
  middleAlign?: boolean;
  /** Renders an `<ol></ol>`. */
  ordered?: boolean;
  /** Wrap horizontal list. */
  wrap?: boolean;
};

export default function List({ children, gutter, horizontal, middleAlign, ordered, wrap }: Props) {
```

stefhatcher:
@schillerk you might need to update your branch with lunar's latest master to get the build to work?

packages/core/src/components/TermList/Term.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export default function Term({ label, after, endAlign, regular, children }: Props) {
+  return (
+    <>
+      <dt>
+        <Row after={endAlign && after}>
+          <Text inline small={!regular}>
```
milesj: I have some designs that uppercases everything. May I get an `uppercased` prop here? :P

packages/core/src/components/TermList/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+};
+
+function TermList({ children, cx, styles }: Props & WithStylesProps) {
+  return <dl className={cx(styles.termList)}>{children}</dl>;
+}
+
+export default withStyles(({ unit, font }) => ({
+  termList: {
+    // Reset dl styles
+    marginBlockStart: 0,
```
milesj: Do these new rules actually compile correctly? (We support more than just chrome)

packages/core/src/components/TermList/Term.tsx
schillerk: I think this is usually worse for readability, but added

schillerk:
Abandoned for https://github.com/airbnb/lunar/pull/297

PR #136, opened 2019-07-23
Ks tooltip hover

schillerk:

## Description
This is kind of messy, basically we want to:
- Let users hover over tooltips without the tooltip closing
- Add an invisible border to increase the hoverable area

Downsides are:
- This removes the Overlay wrapper
- Still not a11y friendly

I can fix tests, but wanted to have the design discussion first.

## Checklist

### DIFF
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -28,7 +28,7 @@ export type Props = {
 
 export type State = {
   labelID: string;
-  hoveringButton: boolean;
+  hoveringTarget: boolean;
   hoveringTooltip: boolean;
   tooltipHeight: number;
   targetRect: ClientRect;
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Tooltip/index.tsx
```diff
@@ -177,25 +184,29 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
...
+        <div aria-labelledby={labelID} className={cx(!disabled && underlined && styles.underlined)}>
           {children}
         </div>
 
         <div id={labelID} className={cx(styles.offscreen)}>
           {content}
         </div>
 
-        <Overlay open={open} onClose={this.handleClose} noBackground>
+        {open && (
```
stefhatcher: if this were done with hover & focus styles, it'd be accessible (e.g. the solution that doesn't use `onMouse` events). the tooltip div would be visually hidden until its parent has hover or focus styles activated. since it is a child of the parent, with focus, should still stay open?

schillerk:
Abandoning to pursue a more a11y friendly Popover

PR #137, opened 2019-07-23
update(Row): Add inline prop.

stefhatcher:

## Description

Adds an inline prop, switching the row from being display flex, to flex-inline. When it's inline, its column spacing is the same as compact (8px).

## Motivation and Context

Need it! Handy for after content to be right up against main content with middle alignment as an option!

## Testing

- added test
- added story

## Screenshots

Screen Shot 2019-07-23 at 4 55 11 PM

## Checklist

### DIFF
--- a/packages/core/src/components/Row/index.tsx
+++ b/packages/core/src/components/Row/index.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import withStyles, { WithStylesProps } from '../../composers/withStyles';
+import { AutoSizer } from 'react-virtualized';
 
 export type Props = {
   /** The contents following the primary contents. */
@@ -62,7 +63,6 @@ export class Row extends React.Component<Props & WithStylesProps> {
         className={cx(
           styles.row,
           { maxHeight },
-          inline && styles.row_inline,
           compact && styles.row_compact,
           spacious && styles.row_spacious,
           middleAlign && styles.row_middleAlign,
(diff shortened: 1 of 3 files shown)


milesj:
`flex-inline` is super buggy in some browsers. That's pretty much why we haven't used it in other places.

stefhatcher:
i've been anit-`inline-flex` in the past, but it's support looks good now and can't be as funky as switching everything to inline-block and v-align just for inlining. inline block introduces a bunch of funk. i could switch it to the even newer `inline-grid` :D

stefhatcher:
it has pretty great global usage now:
Screen Shot 2019-07-24 at 9 43 01 AM

milesj:
It's not so much usage, but broken implementations. For example, all buttons and inline-block elements were `inline-flex` when we originally started this UI, and they were so problematic we had to revert them.

stefhatcher:
checked out what dls does and they use inline-flex as well, but they put it on the children and keep the parent flex. updated to match that! works well. also cleaned up the inner spacing to be consistent with Grid and AdaptiveGrid gap/gutter.

stefhatcher:
bump @milesj

packages/core/src/components/Row/index.tsx
```diff
@@ -65,13 +69,18 @@ export class Row extends React.Component<Props & WithStylesProps> {
...
         )}
       >
-        {before && (
-          <div className={cx(compact ? styles.before_compact : styles.before)}>{before}</div>
-        )}
-
-        <div className={cx(styles.primary, truncated && styles.primary_truncated)}>{children}</div>
-
-        {after && <div className={cx(compact ? styles.after_compact : styles.after)}>{after}</div>}
+        {before && <div className={cx(styles.before, inline && styles.inline)}>{before}</div>}
```
milesj: I don't believe we should remove the compact padding in this PR. Also seems weird to only have 8px padding on spacious (28px) rows.

stefhatcher: the 8px is to get the column spacing to be consistent with other column spacing in Grid and AdaptiveGrid. i can do it in a separate PR or keep it here

milesj: Should go through design approval first.

stefhatcher: talked with design - they are 👍 with this change. will add the larger column style for `spacious` too

stefhatcher:
updates made based on design feedback @milesj

PR #138, opened 2019-07-24
new: Add useQuery, useMutation, and more hooks

milesj:

## Description

Re-exports the official React hooks.

## Motivation and Context

Because hooks are the future.

## Testing

N/A

## Checklist

### DIFF
--- a/packages/apollo/package.json
+++ b/packages/apollo/package.json
@@ -31,6 +31,7 @@
     "react-test-renderer": "^16.8.6"
   },
   "dependencies": {
+    "@apollo/react-hooks": "^0.1.0-beta.11",
     "@types/graphql": "^14.2.2",
     "@types/lodash": "^4.14.134",
     "apollo-cache-inmemory": "^1.6.2",
--- a/packages/apollo/src/hooks/useLazyQuery.ts
+++ b/packages/apollo/src/hooks/useLazyQuery.ts
@@ -1,0 +1,3 @@
+import { useLazyQuery } from '@apollo/react-hooks';
+
+export default useLazyQuery;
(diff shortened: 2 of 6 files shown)


ostrgard:
Hi Miles. I love that you're working with hooks! 

I've played around with this a bit today and it seems that `useQuery` etc. cannot find the client context. The component is wrapped in a `Provider` and the `Query` component does work. When refactoring to `useQuery` I'm getting the following error:

```
Uncaught Invariant Violation: Could not find "client" in the context or passed in as an option. Wrap the root component in an <ApolloProvider>, or pass an ApolloClient instance in via options.
```

Here's a potential fix for the hooks, with `useQuery.ts` as example, that just declares same types as `@apollo/react-hooks/useQuery` and patches the client with `Apollo.getClient()`, like it's used in `Provider`.
```typescript
import { OperationVariables, QueryResult } from '@apollo/react-common';
import { useQuery, QueryHookOptions } from '@apollo/react-hooks';
import { DocumentNode } from 'graphql';
import Apollo from '../..';

export default function<TData = any, TVariables = OperationVariables>(
  query: DocumentNode,
  options: QueryHookOptions<TData, TVariables> = {},
): QueryResult<TData, TVariables> {
  return useQuery(query, { client: Apollo.getClient(), ...options });
}
```
I might have missed some configuration, and if it's working as expected for you, just disregard this comment. There might also be a deeper issue that should be solved. Also, I'm sorry to just throw in a comment here for which I wont be able to see your replies, as I leave for 4 weeks of PTO at the end of the day. 

Hope it's of use. 

Cheers!

milesj:
@ostrgard Thanks for the feedback. It's interesting that it's not finding the context provider, I wonder if it's a reference issue. I'll dig in a bit further.

PR #142, opened 2019-07-25
fix: Override defaultValue and onChange prop types.

milesj:

## Description

Tried the recent form type changes I made, and it seems to collide with `defaultValue` from `@types/react`, as seen here:

```
src/views/Form.tsx:21:13 - error TS2322: Type 'string | null | undefined' is not assignable to type 'string | string[] | (string & string[]) | (string[] & string) | (string[] & false) | (string[] & true) | undefined'.
  Type 'null' is not assignable to type 'string | string[] | (string & string[]) | (string[] & string) | (string[] & false) | (string[] & true) | undefined'.

21             defaultValue={data.name}
               ~~~~~~~~~~~~

  node_modules/@types/react/index.d.ts:1639:9
    1639         defaultValue?: string | string[];
                 ~~~~~~~~~~~~
```

Since we provide our own `defaultValue` (and `onChange`), I updated the types to just omit them.

## Motivation and Context

To fix the types.

## Testing

I verified the built `.d.ts` files.

## Checklist

### DIFF
--- a/packages/forms/src/composers/connectToForm.tsx
+++ b/packages/forms/src/composers/connectToForm.tsx
@@ -25,13 +25,10 @@ export const PROP_NAMES = [
   'validator',
 ];
 
-// Our composer provides implementations for required props
-// in the wrapped component, so we need to mark them as optional.
-export type OnChangeHandler = (...args: any[]) => any;
-
-export type OptionalOnChange<T> = T extends { onChange: OnChangeHandler }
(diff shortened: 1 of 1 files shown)


packages/forms/src/composers/connectToForm.tsx
```diff
@@ -56,6 +53,7 @@ export interface ConnectToFormProps<T> {
 export interface ConnectToFormWrapperProps<T> extends Field<T> {
```
milesj: `defaultValue` comes from `Field`.

PR #144, opened 2019-08-01
update(DataTable): Add autoHeight prop

schillerk:

## Description
There are use cases for DataTable that involve giving it full height. Right now we satisfy these by using something like `showAllRows`, but that removes all benefits of virtualization, and means that scroll is on the page itself instead of the table.

## Checklist


packages/core/src/components/DataTable/index.tsx
```diff
@@ -118,12 +119,13 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
     }
   }
 
-  private getTableHeight = (expandedDataList: ExpandedRow[]) => {
-    const { height, rowHeight, showAllRows } = this.props;
+  private getTableHeight = (expandedDataList: ExpandedRow[]): number => {
+    const { autoHeight, height, rowHeight, showAllRows } = this.props;
```
milesj: `autoHeight` isn't used anywhere?

schillerk: Removed the lint, should work now

milesj:
According to the diffs, this completely removes that header. Intentional?

packages/core/src/components/DataTable/types.ts
```diff
@@ -39,6 +39,8 @@ export type HeaderButton = {
 export type DefaultDataTableProps = keyof DataTableProps;
 
 export interface DataTableProps {
+  /** If enabled height will be infered from parent. */
```
stefhatcher: `infered` -> `inferred`

schillerk:
@milesj not intentional, thanks for catching this, not sure how I missed it on Happo

schillerk:
Requested additional review from DX

packages/core/src/components/DataTable.story.tsx
```diff
@@ -126,35 +126,26 @@ export class SearchDemo extends React.Component {
...
-        />
-      </>
+        <div style={{ flexGrow: 1 }}>
+          <DataTable
+            tableHeaderLabel="My Great Table"
+            data={data}
+            filterData={filteredData}
+            autoHeight
+            expandable
+            columnHeaderHeight="micro"
```
stefhatcher: small nit: move boolean props to the top to be consistent

packages/core/src/components/DataTable/index.tsx
```diff
@@ -370,17 +375,13 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
-        {this.shouldRenderTableHeader() && (
-          <AutoSizer disableHeight>
-            {({ width }: { width: number }) => this.renderTableHeader(width)}
-          </AutoSizer>
-        )}
-        <div className={cx(styles.table_container)}>
-          <AutoSizer disableHeight>
-            {({ width }: { width: number }) => (
+      <div className={cx(styles.table_container)}>
+        <AutoSizer disableHeight={!autoHeight}>
```
kristw: can `autoHeight` be `0` and what does it mean?

packages/core/src/components/DataTable.story.tsx
williaster: there's a lint rule for this isn't there?

milesj: Yeah. Enabled in Nimbus but probably not here.

williaster:
can you add screen shots?

packages/core/src/components/DataTable/index.tsx
```diff
@@ -122,12 +123,14 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
     const { height, rowHeight, showAllRows } = this.props;
 
-    return showAllRows
-      ? expandedDataList.length * getHeight(rowHeight) + this.getColumnHeaderHeight()
-      : height || 0;
+    if (showAllRows) {
+      return expandedDataList.length * getHeight(rowHeight) + this.getColumnHeaderHeight();
+    }
+
+    return height!;
```
isaacpz: Is this non-null assertion always correct? What if no height prop is provided? `getTableHeight` used to default to 0

schillerk: Yeah, there's a `defaultProp` for `height`, but unfortunately typescript can't pick this up.

schillerk:
@williaster Sigh, yeah, there's some weird stuff going on with storybook and DataTable. Conglei's run into it as well. When I walk through manually, data appears for all the stories except "Table with search box". Everything looks normal if you reload the page, but if you're switching from another story the data doesn't render. I'm guessing this is what Happo's running into as well.

packages/core/src/components/DataTable/index.tsx
```diff
@@ -413,6 +413,7 @@ export default withStyles(
   theme => ({
     table_container: {
       overflowX: 'auto',
+      height: '100%',
```
stefhatcher: is this PR still necessary, given this change (adding the height: 100%)? https://github.com/airbnb/lunar/pull/168

schillerk: Just talked to Conglei again, I don't think #168 is the right fix since we're still struggling to reproduce the issue, so I might just abandon that PR

PR #149, opened 2019-08-06
new(helpers): Add typescript helpers

piewell:

Adding useful Typescript helpers to Lunar for all!

As we move to the new front-end architecture, some files cannot be shared between the old and new folders. This makes the file location moot, and gives other apps access to simple Typescript helpers.

### DIFF
--- a/packages/core/helpers/typescript.ts
+++ b/packages/core/helpers/typescript.ts
@@ -1,0 +1,36 @@
+import { Omit } from 'utility-types/dist/mapped-types';
+
+export function filterNulls<T>(value: T | null): value is T {
+  return Boolean(value);
+}
+
+export function hasProps<T, K extends keyof T>(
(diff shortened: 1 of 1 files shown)


piewell:
@milesj Are we OK with putting these helpers in Lunar? I'm having some trouble getting the checks to work, but aside from that it's just a copy/paste

stefhatcher:
did some thinking on this, and i don't think this should be added to lunar. they aren't being used by any of the components in lunar and lunar is primarily a toolkit of presentational, some functional, components, not so much helpers. whether the helpers are being used in old vs new folders internally and this would help with some lint errors, it's an application level concern, not a lunar concern. thoughts @milesj ?

piewell:
Closing pull request - decision made to not move these helpers into Lunar

PR #153, opened 2019-08-08
update(Copy, MenuItem): Allow id to pass through to Link for tracking

alecklandgraf:
to: @milesj @stefhatcher @hayes 

## Description

Allows passing an `id` prop to `Copy`, which is passed through to `Link`

## Motivation and Context

Links and Buttons with an `id` can be tracked easier. 

## Testing

Updated tests. 

## Checklist

### DIFF
--- a/packages/core/src/components/Breadcrumbs/Breadcrumb.tsx
+++ b/packages/core/src/components/Breadcrumbs/Breadcrumb.tsx
@@ -14,6 +14,8 @@ export type Props = {
   highlighted?: boolean;
   /** @ignore */
   horizontal?: boolean;
+  /** Pass an HTML element attribute id */
+  id?: string;
   /** Content to within the Breadcrumb. */
   label: string;
   /** Render an anchor link with a URL instead of a button. */
@@ -48,6 +50,7 @@ class Breadcrumb extends React.Component<Props & WithStylesProps> {
       disabled,
       hideIcon,
       highlighted,
+      id,
       label,
       href,
       onClick,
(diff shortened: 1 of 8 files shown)


alecklandgraf:
@milesj getting an odd type error locally. 

Any idea?

```
Type '{ text: string; id: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Copy> & Readonly<Props> & Readonly<{ children?: ReactNode; }>'.
  Property 'id' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Copy> & Readonly<Props> & Readonly<{ children?: ReactNode; }>'.ts(2322)
```

stefhatcher:
@alecklandgraf i think you'll have to add `id` as a type on ButtonOrLink https://github.com/airbnb/lunar/blob/master/packages/core/src/components/private/ButtonOrLink.tsx

which Link uses https://github.com/airbnb/lunar/blob/master/packages/core/src/components/Link/index.tsx#L10

alecklandgraf:
@stefhatcher @milesj can you advise on how to handle the Happo test issues.

milesj:
While you're at it, maybe we should just pass `id` for any components that wrap `Link`?

alecklandgraf:
@milesj will add few more here and leave out the rest until we need them. Some things are probably better hard coded, like the ZoomControl buttons, but we'd run the risk of duplicating `id` in the DOM multiple zoom controls are used on one page.

milesj:
@alecklandgraf Yeah maybe a `data-tracking-id` or something similar.

PR #154, opened 2019-08-09
update(DataTable): Add data type support to row renderers. Render rows/cells as JSX.

milesj:

## Description

At the moment, the `RendererProps` type for `row.rowData.data` is an any object and not the actual shape of the record. This updates the type to use generics so we can type the renderer correctly.

```ts
function NameCell(props: RendererProps<{ name: string }>) {
  const name = props.row.rowData.name; // string
}
```

This also fixes an issue where React hooks were not usable in the renderers since they weren't rendered as React elements, and instead were called as functions.

## Motivation and Context

Relying on any is problematic.

## Checklist


packages/core/src/components/DataTable/columns/renderDataColumns.tsx
```diff
@@ -63,15 +61,14 @@ export default function renderDataColumns(
...
-          return colSpanRenderer(rendererArguments);
+          return React.createElement(colSpanRenderer, rendererArguments);
         }
       }
     }
 
-    const contents: React.ReactNode = customRenderer
-      ? customRenderer(rendererArguments)
-      : renderDefaultContent(rendererArguments);
+    const contents = React.createElement(customRenderer || DefaultRenderer, rendererArguments);
```
milesj: Fixes hooks.

packages/core/src/components/DataTable/columns/renderDataColumns.tsx
```diff
@@ -43,17 +43,15 @@ export default function renderDataColumns(
...
     const { isChild } = metadata;
     const customRenderer = renderers && renderers[key];
-
     const indentSize = !expandable || !isLeftmost ? 2 : 2.5;
     const spacing = isChild || !((expandable || selectable) && isLeftmost) ? indentSize : 0;
-
     const rendererArguments: RendererProps = {
       row,
-      key,
+      keyName: key,
```
milesj: @schillerk Breaking?

schillerk: Looks like it's only being used in Bighead, not a big deal

PR #156, opened 2019-08-12
update(DataTable): Sort null and undefined last.

schillerk:

## Description
Currently undefined values are treated as small values, and null values don't get any special treatment. Chris Council and @williaster have both asked for this new treatment, with null and undefined sorted last regardless of ascending/descending order.

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/helpers/sortData.ts
+++ b/packages/core/src/components/DataTable/helpers/sortData.ts
@@ -1,20 +1,26 @@
 import { SortDirection, SortDirectionType } from 'react-virtualized';
 import { GenericRow } from '../types';
 
-function sortDesc(a: any, b: any) {
-  if (typeof b === 'undefined' || a < b) {
+// https://stackoverflow.com/questions/29829205/sort-an-array-so-that-null-values-always-come-last
+function sort(a: any, b: any, ascending: boolean = false) {
+  // Equal items sort equally.
+  if (a === b) {
+    return 0;
+  }
+  // null sort after anything else.
+  if (a == null || a == 'undefined') {
     return 1;
   }
 
-  return -1;
-}
-
-function sortAsc(a: any, b: any) {
-  if (typeof b === 'undefined' || a < b) {
+  if (b == null || b == 'undefined') {
     return -1;
   }
-
-  return 1;
+  // If ascending, small sorts first.
+  if (ascending) {
+    return a < b ? -1 : 1;
+  }
+  // If descending, largest. sorts first
+  return a < b ? 1 : -1;
 }
 
 export default function sortData<T extends GenericRow>(
@@ -25,10 +31,10 @@ export default function sortData<T extends GenericRow>(
 ): T[] {
   if (sortBy && keys.includes(sortBy)) {
     if (sortDirection === SortDirection.ASC) {
-      return list.slice().sort((a: T, b: T) => sortAsc(a.data[sortBy], b.data[sortBy]));
+      return list.slice().sort((a: T, b: T) => sort(a.data[sortBy], b.data[sortBy], true));
     }
 
-    return list.slice().sort((a: T, b: T) => sortDesc(a.data[sortBy], b.data[sortBy]));
+    return list.slice().sort((a: T, b: T) => sort(a.data[sortBy], b.data[sortBy]));
   }
 
   return list;
--- a/packages/core/src/components/DataTable/helpers/sortData.ts
+++ b/packages/core/src/components/DataTable/helpers/sortData.ts
@@ -8,11 +8,11 @@ function sort(a: any, b: any, ascending: boolean = false) {
     return 0;
   }
   // null sort after anything else.
-  if (a === null || a === 'undefined') {
+  if (a == null || a === 'undefined') {
     return 1;
   }
 
-  if (b === null || b === 'undefined') {
+  if (b == null || b === 'undefined') {
     return -1;
   }
   // If ascending, small sorts first.
--- a/packages/core/src/components/DataTable/helpers/sortData.ts
+++ b/packages/core/src/components/DataTable/helpers/sortData.ts
@@ -8,11 +8,11 @@ function sort(a: any, b: any, ascending: boolean = false) {
     return 0;
   }
   // null sort after anything else.
-  if (a == null || a === 'undefined') {
+  if (a === null || a === 'undefined') {
     return 1;
   }
 
-  if (b == null || b === 'undefined') {
+  if (b === null || b === 'undefined') {
     return -1;
   }
   // If ascending, small sorts first.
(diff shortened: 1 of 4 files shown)


packages/core/src/components/DataTable/helpers/sortData.ts
```diff
@@ -1,20 +1,26 @@
...
-function sortDesc(a: any, b: any) {
-  if (typeof b === 'undefined' || a < b) {
+// https://stackoverflow.com/questions/29829205/sort-an-array-so-that-null-values-always-come-last
+function sort(a: any, b: any, ascending: boolean = false) {
+  // Equal items sort equally.
+  if (a === b) {
+    return 0;
+  }
+  // null sort after anything else.
+  if (a == null || a == 'undefined') {
```
soboko: Do you mean `typeof a === 'undefined'`?

packages/core/src/components/DataTable/helpers/sortData.ts
```diff
@@ -1,20 +1,26 @@
...
+  if (b == null || b == 'undefined') {
     return -1;
   }
-
-  return 1;
+  // If ascending, small sorts first.
+  if (ascending) {
+    return a < b ? -1 : 1;
+  }
+  // If descending, largest. sorts first
```
soboko: I'd remove the . after largest

packages/core/src/components/DataTable/helpers/sortData.ts
stefhatcher: if `a` is `null` or `undefined` both of these double equals `==` checks return true - redundant. personally, i try to always use triple equals checks. is this truly sorting null to come last over undefined?

Screen Shot 2019-08-12 at 4 23 52 PM

the stackoverflow answer you linked also used `=== null` check

schillerk: I haven't thought about null/undefined sorting relative to each other. Just updated it to use `  if (b == null || b === 'undefined') {`

`=== null` doesn't work in the storybook, 
http://adripofjavascript.com/blog/drips/equals-equals-null-in-javascript.html

stefhatcher: so, to be more explicit with the checks:

`a === null || typeof a === 'undefined'`
`b === null || typeof b === 'undefined'`

stefhatcher: this'll sort `var c = ['foo', null, 'bar', null, undefined, 'baz', undefined]` to `["bar", "baz", "foo", null, undefined]`
(example minus the asc/desc logic):

Screen Shot 2019-08-12 at 4 44 46 PM

is that what you're expecting?

stefhatcher: to check if it is `undefined` you have to do `typeof b ==='undefined'`, doing just `b === 'undefined'` won't return the proper boolean result.

dang it storybook! knowing that it won't all `=== null`, you can reduce the two checks to be just `b == null` (no `typeof b === 'undefined'`)

so just `if (a == null) return 1;` and `if (b == null) return -1`

packages/core/test/components/DataTable.test.tsx
```diff
@@ -118,8 +118,8 @@ const data: ParentRow[] = [
   {
     data: {
       name: 'Frontend Fabien',
-      jobTitle: 'Engineer',
-      tenureDays: 600,
+      jobTitle: undefined,
+      tenureDays: null,
```
stefhatcher: add a test that sorting returns `null` and `undefined` at the end? add a sort test?

packages/core/src/components/DataTable/helpers/sortData.ts
schillerk: Awesome this works, thank you so much for helping me think through this!

packages/core/src/components/DataTable/helpers/sortData.ts
```diff
@@ -25,10 +31,10 @@ export default function sortData<T extends GenericRow>(
 ): T[] {
   if (sortBy && keys.includes(sortBy)) {
     if (sortDirection === SortDirection.ASC) {
-      return list.slice().sort((a: T, b: T) => sortAsc(a.data[sortBy], b.data[sortBy]));
+      return list.slice().sort((a: T, b: T) => sort(a.data[sortBy], b.data[sortBy], true));
     }
 
-    return list.slice().sort((a: T, b: T) => sortDesc(a.data[sortBy], b.data[sortBy]));
+    return list.slice().sort((a: T, b: T) => sort(a.data[sortBy], b.data[sortBy]));
```
kristw: Could keep `sortAsc()` and reverse? 

```ts
return list.slice().sort((a: T, b: T) => -sortAsc(a.data[sortBy], b.data[sortBy]));
```

schillerk: Ohh that's cool, I never knew you could use `-` to reverse.

But it doesn't work here cause then you'd end up with `null` back on top :(

kristw: oh right. the nulls

packages/core/src/components/DataTable/helpers/sortData.ts
```diff
@@ -1,20 +1,27 @@
...
+// https://stackoverflow.com/questions/29829205/sort-an-array-so-that-null-values-always-come-last
+function sort(a: any, b: any, ascending: boolean = false) {
+  // Equal items sort equally.
+  if (a === b) {
+    return 0;
+  }
+  // null sort after anything else.
+  /* eslint-disable no-eq-null */
+  if (a == null) {
     return 1;
```
kristw: This do not distinguish `null` and `undefined` and do not guarantee ordering between `null` and `undefined`.

`[null, undefined, null, 'a', undefined]` should be `['a', null, null, undefined, undefined]`.

packages/core/src/components/DataTable/helpers/sortData.ts
```diff
@@ -1,20 +1,27 @@
...
-  if (typeof b === 'undefined' || a < b) {
+// https://stackoverflow.com/questions/29829205/sort-an-array-so-that-null-values-always-come-last
+function sort(a: any, b: any, ascending: boolean = false) {
+  // Equal items sort equally.
+  if (a === b) {
+    return 0;
+  }
+  // null sort after anything else.
+  /* eslint-disable no-eq-null */
+  if (a == null) {
```
kristw: This does not distinguish `null` and `undefined`.

Does it ensure that `[undefined, null, undefined, null]` will become `[undefined, undefined, null, null]`?

stefhatcher: yeah - we had a back and forth about it. right now, not distinguishing between `null` and `undefined`. it'll sort `null` before `undefined`. so, `[null, null, undefined, undefined]`

PR #159, opened 2019-08-13
update(DataTable): Only render top border when there's a header.

schillerk:

## Description
There's a top border on ColumnHeaders to separate them from the header label, but this is unnecessary when the header label is absent and adds extra visual noise.

## Motivation and Context
Feature request from Ken Chen.

## Screenshots
Screen Shot 2019-08-13 at 4 32 03 PM
Screen Shot 2019-08-13 at 4 32 09 PM

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/TableHeader.tsx
+++ b/packages/core/src/components/DataTable/TableHeader.tsx
@@ -142,5 +142,6 @@ export default withStyles(({ unit, color }) => ({
     justifyContent: 'space-between',
     paddingLeft: 2 * unit,
     paddingRight: 2 * unit,
+    borderBottom: `1px solid ${color.core.neutral[1]}`,
   },
 }))(TableHeader);
--- a/packages/core/src/components/DataTable/TableHeader.tsx
+++ b/packages/core/src/components/DataTable/TableHeader.tsx
@@ -133,7 +133,7 @@ export function TableHeader({
   );
 }
 
-export default withStyles(({ unit, color }) => ({
+export default withStyles(({ unit, color, ui }) => ({
   tableHeader_inner: {
     background: color.accent.bg,
     display: 'flex',
@@ -142,6 +142,6 @@ export default withStyles(({ unit, color }) => ({
     justifyContent: 'space-between',
     paddingLeft: 2 * unit,
     paddingRight: 2 * unit,
-    borderBottom: `1px solid ${color.core.neutral[1]}`,
+    borderBottom: ui.border,
   },
 }))(TableHeader);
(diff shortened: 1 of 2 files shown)


packages/core/src/components/DataTable/TableHeader.tsx
```diff
@@ -142,5 +142,6 @@ export default withStyles(({ unit, color }) => ({
     justifyContent: 'space-between',
     paddingLeft: 2 * unit,
     paddingRight: 2 * unit,
+    borderBottom: `1px solid ${color.core.neutral[1]}`,
```
stefhatcher: `ui.border`?

schillerk: No :( `ui.border` is `color.neutral[2]`

milesj: Any reason why this border is different?

stefhatcher: it'd be great if it were using the same border color as the rest of lunar

packages/core/src/components/DataTable/index.tsx
```diff
@@ -410,22 +410,19 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
-  theme => ({
+  ({ color }) => ({
     table_container: {
       overflowX: 'auto',
     },
     column_header: {
-      borderTop: '1px solid',
-      borderBottom: '1px solid',
-      borderColor: theme!.color.core.neutral[1],
+      borderBottom: `1px solid ${color.core.neutral[1]}`,
```
stefhatcher: `ui.border` here too. the jump from 1 to 2 isn't that large? and would be consistent with the rest of lunar borders, especially in tables

kenchendesign:
+1 on standardizing all the border colors. They should all be the same gray. ui.border is using neutral[2] which has better contrast so I'm down for this choice. Thanks for flagging this @stefhatcher 

I also quickly looked around and found that `Alert` looks like it's not using the right border color either:
image

A separate PR to change that as well?

milesj:
@kenchendesign `Alert` was the outlier because of the box shadow.

kenchendesign:
> @kenchendesign `Alert` was the outlier because of the box shadow.

But cards have the same border color while having box-shadow as well.

schillerk:
Make the switch to `ui.border`

packages/core/src/components/DataTable/index.tsx
```diff
@@ -410,22 +410,19 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
 }
 
 export default withStyles(
-  theme => ({
+  ({ color, ui }) => ({
```
stefhatcher: `color` is no longer needed

PR #163, opened 2019-08-14
update(Copy, Chip, Breadcrumb, MenuItem): Add data-tracking-name

alecklandgraf:

## Description

Adds `trackingName` which will follow our convention of adding a `data-tracking-name` attribute on button and anchor elements. 

## Motivation and Context

`id` should be unique across the page. When `id` is used in reusable components, it can be difficult or impossible for consumers to provide a unique `id`. Adding `data-tracking-name` is a better alternative for tracking. 

https://www.w3.org/TR/html4/struct/global.html#h-7.5.2

Future PRs can add `trackingName` and `data-tracking-name` as needed to provide tracking context for analytics. 

## Testing

Updated tests. 

## Checklist

### DIFF
--- a/packages/core/src/components/Breadcrumbs/Breadcrumb.tsx
+++ b/packages/core/src/components/Breadcrumbs/Breadcrumb.tsx
@@ -24,7 +24,7 @@ export type Props = {
   onClick?: () => void;
   /** Mark the breadcrumb as selected. */
   selected?: boolean;
-  /** Adds a data-tracking-name attribute. */
+  /** A tracking name to identify this component. */
   trackingName?: string;
 };
 
@@ -88,7 +88,7 @@ class Breadcrumb extends React.Component<Props & WithStylesProps> {
           }
           disabled={disabled}
           href={href}
-          data-tracking-name={trackingName}
+          trackingName={trackingName}
           id={id}
           onClick={this.handleClick}
         >
--- a/packages/core/src/components/Breadcrumbs/Breadcrumb.tsx
+++ b/packages/core/src/components/Breadcrumbs/Breadcrumb.tsx
@@ -14,7 +14,7 @@ export type Props = {
   highlighted?: boolean;
   /** @ignore */
   horizontal?: boolean;
-  /** Pass an HTML element attribute id */
+  /** Pass an HTML element attribute id. */
   id?: string;
   /** Content to within the Breadcrumb. */
   label: string;
@@ -24,7 +24,7 @@ export type Props = {
   onClick?: () => void;
   /** Mark the breadcrumb as selected. */
   selected?: boolean;
-  /** Adds a data-tracking-name attribute */
+  /** Adds a data-tracking-name attribute. */
   trackingName?: string;
 };
 
--- a/packages/core/src/components/Breadcrumbs/Breadcrumb.tsx
+++ b/packages/core/src/components/Breadcrumbs/Breadcrumb.tsx
@@ -24,6 +24,8 @@ export type Props = {
   onClick?: () => void;
   /** Mark the breadcrumb as selected. */
   selected?: boolean;
+  /** Adds a data-tracking-name attribute */
+  trackingName?: string;
 };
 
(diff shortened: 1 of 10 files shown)


alecklandgraf:
cc @Alphy11

packages/core/src/components/Breadcrumbs/Breadcrumb.tsx
```diff
@@ -24,6 +24,8 @@ export type Props = {
   onClick?: () => void;
   /** Mark the breadcrumb as selected. */
   selected?: boolean;
+  /** Adds a data-tracking-name attribute */
```
stefhatcher: small nit for these prop descriptions throughout the pr: end with a `.` to be consistent

packages/core/src/components/Copy/index.tsx
```diff
@@ -54,10 +56,10 @@ export default class Copy extends React.Component<Props, State> {
   };
 
   render() {
-    const { prompt, children, id, underlined } = this.props;
+    const { prompt, children, id, trackingName, underlined } = this.props;
     const element = children || (
       // eslint-disable-next-line jsx-a11y/anchor-is-valid
-      <Link id={id}>
+      <Link data-tracking-name={trackingName} id={id}>
```
milesj: `trackingName={trackingName}`? Both `Link` and `Button` should support it.

stefhatcher: agreed - will need to at it to ButtonOrLink https://github.com/airbnb/lunar/blob/master/packages/core/src/components/private/ButtonOrLink.tsx#L7

alecklandgraf: @stefhatcher indeed I added to ButtonOrLink to make the prop available across all the ButtonOrLink consumers.

packages/core/src/components/private/ButtonOrLink.tsx
```diff
@@ -28,6 +28,8 @@ export type Props = {
   openInNewWindow?: boolean;
   /** Rel attribute override for if the component has an href */
   rel?: string;
+  /** Add a data-tracking-name attribute. */
+  trackingName?: string;
```
Alphy11: So in our repo we currently wrap button and link with another component to require a tracking name.  I was wondering if it would make sense to make this required?  I can see arguments both ways though since you aren't required to use tracking with lunar.  It just feels like for airbnb use case this is the best practice.

alecklandgraf: I think it'd break too much to require this as there are so many consumers

alecklandgraf:
@stefhatcher @milesj any idea what this error is? Seems odd

```
Summary of all failing tests
 FAIL  packages/forms/test/components/Form/Multicomplete.test.tsx (445 MB heap size)
  ● <Multicomplete /> › connects to the form
    Warning: React does not recognize the `%s` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `%s` instead. If you accidentally passed it from a parent component, remove it from the DOM element.%s
      4 | 
      5 | console.warn = console.error = function mockedConsole(message) {
    > 6 |   throw new Error(message);
        |         ^
      7 | };
      8 | 
      9 | // ENZYME

```

alecklandgraf:
> ```
> Summary of all failing tests
>  FAIL  packages/forms/test/components/Form/Multicomplete.test.tsx (445 MB heap size)
>   ● <Multicomplete /> › connects to the form
>     Warning: React does not recognize the `%s` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `%s` instead. If you accidentally passed it from a parent component, remove it from the DOM element.%s
>       4 | 
>       5 | console.warn = console.error = function mockedConsole(message) {
>     > 6 |   throw new Error(message);
>         |         ^
>       7 | };
>       8 | 
>       9 | // ENZYME
> ```

I needed to update the `Chip` component which `Multicomplete` was using. Chip can render a `button` or `div` and neither take a `trackingName` attribute.

PR #168, opened 2019-08-20
update(DataTable): set display flex to avoid rare header height bug

schillerk:

## Description
Conglei is running into a bug in superset UI where column headers take up the full width. We investigated, and were unable to discover a cause, but this is a known fix.

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/index.tsx
+++ b/packages/core/src/components/DataTable/index.tsx
@@ -413,6 +413,7 @@ export default withStyles(
   theme => ({
     table_container: {
       overflowX: 'auto',
+      display: 'flex',
     },
     column_header: {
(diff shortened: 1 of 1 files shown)


stefhatcher:
i'm confused - the PR description was a "width" issue, but @conglei says it was a height issue?

stefhatcher:
can you add a story that kind of mocks the bug up and confirms the fix? would be helpful for happo to have

kristw:
@stefhatcher I think Kyle made a typo in PR description. I can confirm that it is a height issue.

schillerk:
Still unable to reproduce, find an underlying cause or identify a robust fix. This only seems to be happening in one specific case so we'll fix that as a one off.

PR #173, opened 2019-08-28
update(DataTable): Add `dynamicRowHeight` prop

schillerk:

## Description
We've been planning to add this support for a while, but @williaster has a more urgent use case for Metrics Explorer and Conglei needs this for Superset UI.

This change allows rows to resize automatically based on the height of their content.

Based on this example from react-virtualized
https://github.com/bvaughn/react-virtualized/blob/master/source/CellMeasurer/CellMeasurer.DynamicHeightTableColumn.example.js

There are some difficulties worth mentioning:
- To enable dynamicRowHeight and showAllRows together we have to add together everything in the cache
- We have to clear the entire cache on sorting and expansion, there ought to be a way to index the row -> height mapping more persistently, but React Virtualized relies on the index to be the actual index in the table, so I don't think this is possible
- Scroll position resets to 0 on row expansion, I've tried various hacks around this (reseting scroll manually), but nothing has really worked
- Filter data is a functional prop that often relies on a changing variable in its closure, so it's difficult to detect changes that results in cache invalidation. Right now we just filter the data in `componentDidUpdate` and compare it to the old filtered data, which is unnecessarily costly. A good future performance fix would be to at least cache the filtered Data.

## Screenshots
Screen Shot 2019-10-07 at 10 04 22 AM
Screen Shot 2019-10-07 at 10 05 15 AM

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/columns/renderDataColumns.tsx
+++ b/packages/core/src/components/DataTable/columns/renderDataColumns.tsx
@@ -21,7 +21,6 @@ type ArgumentsFromProps = {
   styles: WithStylesProps["styles"];
   renderers?: DataTableProps["renderers"];
   zebra?: boolean;
-  rowHeight?: HeightOptions;
   theme?: WithStylesProps["theme"];
   selectable?: boolean;
   expandable?: boolean;
--- a/packages/core/src/components/DataTable/columns/renderDataColumns.tsx
+++ b/packages/core/src/components/DataTable/columns/renderDataColumns.tsx
@@ -44,26 +44,6 @@ export default function renderDataColumns<T>(
     theme
   }: ArgumentsFromProps
 ) {
-  const columnCellRenderer = (columnIdx: number) => (row: VirtualRow) => {
-    const { dataKey, parent, rowIndex } = row;
-
-    const content = renderCell(dataKey, columnIdx, row);
-
-    return (
-      <div className={cx(styles.rowContainer)}>
-        <CellMeasurer
-          cache={cache}
-          columnIndex={columnIdx}
-          key={dataKey}
-          parent={parent}
-          rowIndex={rowIndex}
-        >
-          <div className={cx(styles.row)}>{content}</div>
-        </CellMeasurer>
-      </div>
-    );
-  };
-
   const renderCell = (key: string, columnIndex: number, row: VirtualRow) => {
     const { metadata } = row.rowData;
     const { isChild } = metadata;
@@ -100,6 +80,27 @@ export default function renderDataColumns<T>(
       <Spacing left={spacing} right={2}>
         {contents || ""}
       </Spacing>
+    );
+  };
+
+  const columnCellRenderer = (columnIdx: number) => (row: VirtualRow) => {
+    const { dataKey, parent, rowIndex } = row;
+
+    const content = renderCell(dataKey, columnIdx, row);
+
+    return (
+      <div className={cx(styles.rowContainer)}>
+        <CellMeasurer
+          cache={cache}
+          columnIndex={columnIdx}
+          key={dataKey}
+          //@ts-ignore
+          parent={parent}
+          rowIndex={rowIndex}
+        >
+          <div className={cx(styles.row)}>{content}</div>
+        </CellMeasurer>
+      </div>
     );
   };
 
--- a/packages/core/src/components/DataTable/columns/renderDataColumns.tsx
+++ b/packages/core/src/components/DataTable/columns/renderDataColumns.tsx
@@ -23,6 +23,7 @@ type ArgumentsFromProps = {
   theme?: WithStylesProps['theme'];
   selectable?: boolean;
   expandable?: boolean;
+  dynamicRowHeight?: boolean;
 };
 
 export default function renderDataColumns<T>(
@@ -37,6 +38,7 @@ export default function renderDataColumns<T>(
     selectable,
     showColumnDividers,
     zebra,
+    dynamicRowHeight,
     cx,
     styles,
     theme,
@@ -81,6 +83,14 @@ export default function renderDataColumns<T>(
     const { dataKey, parent, rowIndex } = row;
 
     const content = renderCell(dataKey, columnIdx, row);
+
+    if (!dynamicRowHeight) {
+      return (
+        <div className={cx(styles.rowContainer)}>
+          <div className={cx(styles.row)}>{content}</div>
+        </div>
+      );
+    }
 
     return (
       <div className={cx(styles.rowContainer)}>
--- a/packages/core/src/components/DataTable/columns/renderDataColumns.tsx
+++ b/packages/core/src/components/DataTable/columns/renderDataColumns.tsx
@@ -1,26 +1,26 @@
-import React from "react";
-import { Column, CellMeasurer } from "react-virtualized";
-import DefaultRenderer from "../DefaultRenderer";
-import Spacing from "../../Spacing";
+import React from 'react';
+import { Column, CellMeasurer, CellMeasurerCache } from 'react-virtualized';
+import DefaultRenderer from '../DefaultRenderer';
+import Spacing from '../../Spacing';
 import {
   ColumnMetadata,
   DataTableProps,
   VirtualRow,
   EditCallback,
   WidthProperties,
-  RendererProps
-} from "../types";
-import { WithStylesProps } from "../../../composers/withStyles";
-import { DEFAULT_WIDTH_PROPERTIES } from "../constants";
+  RendererProps,
+} from '../types';
+import { WithStylesProps } from '../../../composers/withStyles';
+import { DEFAULT_WIDTH_PROPERTIES } from '../constants';
 
 type ArgumentsFromProps = {
+  cx: WithStylesProps['cx'];
+  styles: WithStylesProps['styles'];
   columnMetadata?: ColumnMetadata;
   showColumnDividers?: boolean;
-  cx: WithStylesProps["cx"];
-  styles: WithStylesProps["styles"];
-  renderers?: DataTableProps["renderers"];
+  renderers?: DataTableProps['renderers'];
   zebra?: boolean;
-  theme?: WithStylesProps["theme"];
+  theme?: WithStylesProps['theme'];
   selectable?: boolean;
   expandable?: boolean;
 };
@@ -28,8 +28,8 @@ type ArgumentsFromProps = {
 export default function renderDataColumns<T>(
   keys: string[],
   editMode: boolean,
-  onEdit: EditCallback,
-  cache: any,
+  onEdit: EditCallback<T>,
+  cache: CellMeasurerCache,
   {
     columnMetadata,
     expandable,
@@ -39,24 +39,23 @@ export default function renderDataColumns<T>(
     zebra,
     cx,
     styles,
-    theme
-  }: ArgumentsFromProps
+    theme,
+  }: ArgumentsFromProps,
 ) {
-  const renderCell = (key: string, columnIndex: number, row: VirtualRow) => {
+  const renderCell = (key: string, columnIndex: number, row: VirtualRow<T>) => {
     const { metadata } = row.rowData;
     const { isChild } = metadata;
     const customRenderer = renderers && renderers[key];
     const isLeftmost = columnIndex === 0;
     const indentSize = !expandable || !isLeftmost ? 2 : 2.5;
-    const spacing =
-      isChild || !((expandable || selectable) && isLeftmost) ? indentSize : 0;
+    const spacing = isChild || !((expandable || selectable) && isLeftmost) ? indentSize : 0;
     const rendererArguments: RendererProps<T> = {
       row,
       keyName: key as keyof T,
-      editMode,
       onEdit,
       zebra: zebra || false,
-      theme
+      editMode,
+      theme,
     };
 
     if (metadata && metadata.colSpanKey && renderers) {
@@ -69,14 +68,11 @@ export default function renderDataColumns<T>(
       }
     }
 
-    const contents = React.createElement(
-      customRenderer || DefaultRenderer,
-      rendererArguments
-    );
+    const contents = React.createElement(customRenderer || DefaultRenderer, rendererArguments);
 
     return (
       <Spacing left={spacing} right={2}>
-        {contents || ""}
+        {contents || ''}
       </Spacing>
     );
   };
@@ -89,9 +85,9 @@ export default function renderDataColumns<T>(
     return (
       <div className={cx(styles.rowContainer)}>
         <CellMeasurer
+          key={dataKey}
           cache={cache}
           columnIndex={columnIdx}
-          key={dataKey}
           // @ts-ignore We need to pass in the parent node
           parent={parent}
           rowIndex={rowIndex}
@@ -103,13 +99,7 @@ export default function renderDataColumns<T>(
   };
 
   return keys.map((key, idx: number) => {
-    const widthPropertiesOptions = [
-      "maxWidth",
-      "minWidth",
-      "width",
-      "flexGrow",
-      "flexShrink"
-    ];
+    const widthPropertiesOptions = ['maxWidth', 'minWidth', 'width', 'flexGrow', 'flexShrink'];
     const widthProperties: WidthProperties = {};
     widthPropertiesOptions.forEach(property => {
       widthProperties[property] =
@@ -135,7 +125,7 @@ export default function renderDataColumns<T>(
         cellRenderer={columnCellRenderer(idx)}
         className={cx(
           styles && styles.column,
-          showColumnDividers && !isRightmost && styles && styles.column_divider
+          showColumnDividers && !isRightmost && styles && styles.column_divider,
         )}
       />
     );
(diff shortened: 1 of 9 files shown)


packages/core/src/components/DataTable.story.tsx
```diff
@@ -144,8 +166,10 @@ export class SearchDemo extends React.Component {
           </Spacing>
           <div style={{ flexGrow: 1 }}>
             <DataTable
+              renderers={renderers}
               autoHeight
```
williaster: I wonder if it'd be worth changing `autoHeight` to `autoTableHeight` to distinguish this from row height 🤔

packages/core/src/components/DataTable/columns/renderDataColumns.tsx
```diff
@@ -71,12 +72,29 @@ export default function renderDataColumns(
...
+
+    const content = renderCell(dataKey, columnIdx, row);
+
+    return (
+      <div className={cx(styles.rowContainer)}>
+        <CellMeasurer
+          cache={cache}
+          columnIndex={columnIdx}
+          key={dataKey}
+          //@ts-ignore
```
williaster: if this is necessary / stays around, can we add a comment for why it's needed?

packages/core/src/components/DataTable/index.tsx
```diff
@@ -1,6 +1,14 @@
 import React from 'react';
-import { AutoSizer, SortDirection, SortDirectionType, Table } from 'react-virtualized';
+import {
+  CellMeasurerCache,
+  AutoSizer,
+  SortDirection,
+  SortDirectionType,
+  Table,
+} from 'react-virtualized';
+import 'react-virtualized/styles.css';
```
williaster: does the build currently support importing `.css`?

packages/core/src/components/DataTable/index.tsx
```diff
@@ -182,6 +193,7 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
     event: React.SyntheticEvent<EventTarget>,
   ) => {
     event.stopPropagation();
+    // this.cache.clearAll();
```
williaster: is it possible to clear a single row's cache here?

packages/core/src/components/DataTable/index.tsx
```diff
@@ -453,13 +479,13 @@ export default withStyles(
     column_divider: {
       borderRight: ui.border,
     },
-    row: {
+    rowContainer: {
       height: '100%',
       display: 'flex',
```
williaster: have you tried removing `flexbox` on the row container to see if it fixes the `auto` row height calculation? the `react-virtualized` examples does not have this so maybe worth trying.

packages/core/src/components/DataTable.story.tsx
stefhatcher: renaming it would be a breaking change

packages/core/src/components/DataTable.story.tsx
```diff
@@ -144,8 +166,10 @@ export class SearchDemo extends React.Component {
           </Spacing>
           <div style={{ flexGrow: 1 }}>
             <DataTable
+              renderers={renderers}
```
stefhatcher: move this below all boolean props

milesj:
@schillerk Can you rebase? I made some changes to `DataTable`.

packages/core/src/components/DataTable/index.tsx
```diff
@@ -128,8 +136,30 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
+
+    if (
+      filteredData.length > 0 &&
+      (filteredData.length !== oldFilteredData.length ||
+        filteredData.some(
+          (x: IndexedParentRow, i: number) =>
+            x.metadata.originalIndex !== oldFilteredData[i].metadata.originalIndex,
+        ))
+    ) {
+      console.log('123');
```
milesj: Remove logs.

packages/core/src/components/DataTable/index.tsx
```diff
@@ -128,8 +136,30 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
+      setTimeout(() => {
```
milesj: Whats the timeout for?

kristw:
@stefhatcher @milesj @hayes @alecklandgraf Could I ask for a review on this PR, please? Any help would be greatly appreciated. This feature is a long-awaited feature that two high priority projects (namely: Superset and Metric Explorer) are blocked on for further development.

packages/core/src/components/DataTable/index.tsx
```diff
@@ -140,10 +170,24 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
-      return expandedDataList.length * getHeight(rowHeight) + this.getColumnHeaderHeight();
+      if (dynamicRowHeight) {
+        if (Object.values(rowHeights).length > 0) {
+          const totalHeight = Object.values(rowHeights).reduce(
+            (sum: number, measuredRowHeight: number) => sum + measuredRowHeight,
+            0,
+          );
+          return totalHeight + this.getColumnHeaderHeight();
+        }
+        this.setState(prevState => ({ ...prevState }));
```
milesj: `this.forceUpdate()`?

milesj:
Also, please don't Happo approve your own stuff :P

packages/core/src/components/DataTable/index.tsx
williaster: I think this was important for a race condition, if it's needed should def. add a comment given how difficult it was to debug.

packages/core/src/components/DataTable/index.tsx
```diff
@@ -128,8 +136,30 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
+    ) {
+      console.log('123');
+      setTimeout(() => {
+        this.cache.clearAll();
+        this.setState(previousState => ({ ...previousState }));
+      }, 0);
+    }
+
     if (this.props.data !== prevProps.data) {
+      console.log('456');
```
williaster: ditto remove logs

packages/core/src/components/DataTable/index.tsx
```diff
@@ -140,10 +170,24 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
   }
 
   private getTableHeight = (expandedDataList: ExpandedRow[]): number => {
-    const { height, rowHeight, showAllRows } = this.props;
+    const { height, rowHeight, showAllRows, dynamicRowHeight } = this.props;
+    // @ts-ignore _rowHeightCache is missing from DataTable types
+    // eslint-disable-next-line no-underscore-dangle
+    const rowHeights: { [key: number]: number } = this.cache._rowHeightCache;
```
williaster: is `key` really a `number`? could maybe call it `rowIndex` if so.

williaster: could possibly move this inside the relevant `if`/`else` block.

packages/core/src/components/DataTable/index.tsx
```diff
@@ -179,22 +224,27 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
-        newExpandedRows.add(newExpandedRowIndex);
+        newExpandedRows.add(newExpandedRowOriginalIndex);
       }
 
       return {
         expandedRows: newExpandedRows,
       };
     });
+
+    setTimeout(() => {
```
williaster: could also add a comment here for the timeout/race condition, and maybe use `forceUpdate` as miles said above (I forget if that worked or not).

packages/core/src/components/DataTable/index.tsx
```diff
@@ -403,40 +459,39 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
+                  rowCount={expandedData.length}
+                  rowGetter={this.rowGetter(expandedData)}
+                  sort={this.sort}
+                  sortBy={sortBy}
+                  sortDirection={sortDirection}
+                  headerHeight={this.getColumnHeaderHeight()}
+                  headerRowRenderer={ColumnLabels(this.props)}
+                  rowHeight={dynamicRowHeight ? this.cache.rowHeight : HEIGHT_TO_PX[rowHeight!]}
+                  rowStyle={this.getRowStyle(expandedData)}
+                  overscanRowCount={100}
```
williaster: do we need this big of an overscan?

packages/core/src/components/DataTable/index.tsx
```diff
@@ -403,40 +459,39 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
 
     return (
       <AutoSizer disableHeight={!autoHeight}>
-        {({ height, width }: { height: number; width: number }) => (
-          <>
-            {this.shouldRenderTableHeader() && this.renderTableHeader(width)}
-            <div className={cx(styles.table_container, { width })}>
-              <Table
-                ref={propagateRef}
```
williaster: can we add this back?

packages/core/src/components/DataTable/index.tsx
```diff
@@ -370,19 +420,25 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
   rowGetter = (expandedDataList: ExpandedRow[]) => ({ index }: { index: number }) =>
     expandedDataList[index];
 
+  cache = new CellMeasurerCache({
+    fixedHeight: false,
+    fixedWidth: true,
+    defaultHeight: 1,
```
williaster: should this be closer to reasonable height values? like `16` or so?

packages/core/src/components/DataTable/columns/renderDataColumns.tsx
```diff
@@ -71,12 +71,29 @@ export default function renderDataColumns<T>(
...
+  };
+
+  const columnCellRenderer = (columnIdx: number) => (row: VirtualRow<T>) => {
+    const { dataKey, parent, rowIndex } = row;
+
+    const content = renderCell(dataKey, columnIdx, row);
+
+    return (
+      <div className={cx(styles.rowContainer)}>
+        <CellMeasurer
```
williaster: if we are not using `dynamicRowHeight`, do you think we should still use this? or not really a problem?

packages/core/src/components/DataTable/index.tsx
schillerk: Yeah, basically react-virutalized tries to measure rows and populate the cache, but React is trying to re-render at the same time, and we want to make sure the measurements are done before React renders the new table.

packages/core/src/components/DataTable/index.tsx
schillerk: Yeah, wasn't sure which one was worse, changed them all over.

packages/core/src/components/DataTable/index.tsx
schillerk: Yeah, there was some point in debugging where this helped, but now I can't tell the difference between setting it at all. Going to leave it at 16.

packages/core/src/components/DataTable/columns/renderDataColumns.tsx
schillerk: Good catch, doesn't seem like a problem, but probably better for performance.

packages/core/src/components/DataTable/index.tsx
schillerk: Good catch, we can't use `showAllRows` with `dynamicRowHeight` without this, but it's unnecessary otherwise. Changing to
`overscanRowCount={dynamicRowHeight && showAllRows ? expandedData.length : 2}`

schillerk:
@conglei Replied to comments, there are still some diffs on Happo, but a lot of that is just Happo/DataTable bugs, it looks okay on Storybook unless there's a particular item that seems wrong. Got rid of `dynamicRowHeight` on the original `showAllRows` example.

conglei:
Thanks @schillerk for the PR. For the `showAllRows` example, does it mean if we enable 'dynamicRowHeight`, there won't be any paddings by default?

schillerk:
@conglei Yeah that's right, there's actually never any default padding on the table at all. The easiest solution is to pass in a renderer that wraps your content in padding. In the future, we could support:
- a `defaultRenderer` prop
- a `minRowHeight` prop

conglei:
Thanks @schillerk  for the explanation!  I don't think it will be a blocker for this PR.

packages/core/src/components/DataTable/index.tsx
stefhatcher: looks like some `console.log`s are still lingering

packages/core/src/components/DataTable/index.tsx
stefhatcher: remove `console.log`

schillerk: got it!

packages/core/src/components/DataTable/index.tsx
milesj: Maybe overscan be the number of rows currently displayed? Or maybe half?

schillerk: This is possible, but pretty tricky to implement with `dynamicRowHeight`, not sure it's necessary normally since there used to be 0 overscan

schillerk:
@milesj Pretty sure the diffs are a Happo bug, the same stories look fine on Storybook:
Screen Shot 2019-10-08 at 3 40 39 PM

Is there anything else blocking Merge?

packages/core/src/components/DataTable/index.tsx
williaster: @schillerk can you add a comment as discussed?

packages/core/src/components/DataTable/index.tsx
williaster: @schillerk what's the resolution on the type?

packages/core/src/components/DataTable/index.tsx
williaster: @schillerk can you add a comment please?

packages/core/src/components/DataTable/index.tsx
```diff
@@ -370,19 +418,26 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
   rowGetter = (expandedDataList: ExpandedRow[]) => ({ index }: { index: number }) =>
     expandedDataList[index];
 
+  cache = new CellMeasurerCache({
+    fixedHeight: false,
+    fixedWidth: true,
+    defaultHeight: 16,
```
williaster: In thinking about how this will all be used in practice, I think `defaultHeight` and `minHeight` are going to be needed in order to have more pixel-perfect control over the dynamic heights.

can you add these as props with defaults? something like `defaultDynamicRowHeight: 16, minimumDynamicRowHeight: undefined`. I can add this in a follow up PR but I think it'd be nice to save a review cycle.

packages/core/src/components/DataTable/index.tsx
milesj: Yeah no worries then.

packages/core/src/components/DataTable/index.tsx
schillerk: Added

packages/core/src/components/DataTable/types.ts
```diff
@@ -51,8 +51,12 @@ export interface DataTableProps {
   columnMetadata?: ColumnMetadata;
   /** Array of data rows. */
   data?: ParentRow[];
+  /** If dynamicRowHeight is enabled, thiis sets the default value for measured row height. */
```
stefhatcher: `thiis` -> `this` in a few spots

schillerk: got it

packages/core/src/components/DataTable/index.tsx
```diff
@@ -45,7 +52,9 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
     columnMetadata: {},
     columnToLabel: {},
     data: [],
+    defaultDynamicRowHeight: 16,
```
williaster: thanks for adding these!

packages/core/src/components/DataTable/index.tsx
```diff
@@ -128,7 +138,28 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
...
+    if (
+      filteredData.length > 0 &&
+      (filteredData.length !== oldFilteredData.length ||
+        filteredData.some(
+          (x: IndexedParentRow, i: number) =>
+            x.metadata.originalIndex !== oldFilteredData[i].metadata.originalIndex,
+        ))
+    ) {
+      // We need to make sure the cache is cleared before React tries to re-render.
+      setTimeout(() => {
```
williaster: I guess technically this only needs to be called if `dynamicRowHeight === true`?

packages/core/src/components/DataTable/index.tsx
williaster: ditto only need to call this if `dynamicRowHeight === true`

packages/core/src/components/DataTable/index.tsx
stefhatcher: yeah - add a check for `dynamicRowHeight` as the first condition in the `if` statement

packages/core/src/components/DataTable/index.tsx
stefhatcher: ditto - wrap around with an `if (dynamicRowHeight)`

PR #175, opened 2019-08-29
fix(FormField): Fix CSS typo.

davidchang:

## Description

Stumbled upon this CSS-in-JS typo related to `alignItems`. Not really sure where this is actually used (my guess is that it isn't), but putting in this PR to draw attention to it.

## Motivation and Context

Correctness. Though, arguably, if no one has noticed it by now, maybe it doesn't make a difference and the rule ought to be removed. I think it'd have to be a case where you want both an `inline` and a `topAlign`ed FormField - but i don't see how to actually pass in topAlign to a FormField, since it's not part of FormField's Props or properly assigned in `partitionFieldProps`.

## Testing

TODO

## Screenshots

TODO

## Checklist

### DIFF
--- a/packages/core/src/components/FormField/index.tsx
+++ b/packages/core/src/components/FormField/index.tsx
@@ -202,7 +202,7 @@ export default withStyles(({ unit }) => ({
   },
 
   content_topAlign: {
-    alignItems: 'flext-start',
+    alignItems: 'flex-start',
   },
 
   label: {


davidchang:
thanks @stefhatcher ! actually, can you merge this ? i don't have write access

stefhatcher:
i will once happo stops having issues :(

PR #176, opened 2019-08-29
update(RadioButton): support small prop and make children full width

davidchang:

## Description

2 changes:
- support `small` sizing for RadioButton
- make RadioButton children width: 100%.

## Motivation and Context

I have a use case for smaller radio buttons than the current large radio buttons. I was surprised that this was not already supported, as I believe it is already supported for other form fields.

I also extended RadioButton children to ensure it's 100%, as the designs I currently have call for the label to be left-aligned and then something else to be right-aligned (an indicator of a keyboard shortcut that can be used to select the value instead)

## Testing

See screenshots from Storybook below.

## Screenshots

new story in storybook:

image

using a Row with something in the after also shows that the change for width:100% works as intended.

large story, unchanged, for reference:

image

## Checklist

### DIFF
--- a/packages/core/src/components/RadioButton.story.tsx
+++ b/packages/core/src/components/RadioButton.story.tsx
@@ -1,7 +1,10 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
 import { action } from '@storybook/addon-actions';
+import lunar from ':storybook/images/lunar-logo.png';
 import RadioButton from './RadioButton';
+import Row from './Row';
+import ProfilePhoto from './ProfilePhoto';
 
 storiesOf('Core/RadioButton', module)
   .addParameters({
@@ -86,7 +89,7 @@ storiesOf('Core/RadioButton', module)
       />
     </>
   ))
-  .add('As a large clickable button.', () => (
+  .add('As a clickable button.', () => (
     <RadioButton
       name="radio-basic"
       label="Label"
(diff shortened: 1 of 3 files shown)


milesj:
`small` and `large` already mean something in regards to forms. Maybe `compact` makes more sense here?

davidchang:
ahh - i thought `small` was being introduced to be able to deprecate the usage of `compact` (i can't remeber how i got that idea... maybe from reading through the changelog?). i'll rename to compact.

milesj:
@davidchang It is, but it feels kind of weird for `small` to only affect the button form. If we ever add small/large radios/checks themselves, then this prop would be invalid.

davidchang:
PTAL @milesj

packages/core/src/components/RadioButton.story.tsx
```diff
@@ -86,13 +89,27 @@ storiesOf('Core/RadioButton', module)
...
       name="radio-basic"
       label="Label"
       labelDescription="This is a label description."
       value="foo"
       onChange={action('onChange')}
       button
+      topAlign
     />
+  ))
+  .add('As a small clickable button.', () => (
```
stefhatcher: update this description to be `As a compact, clickable button.`

stefhatcher:
might need to update with master again? lots of unexpected happo diffs

davidchang:
🤷‍♂ i believe i am rebased onto master? @stefhatcher

milesj:
@davidchang Can you make the same changes to `CheckBox`?

packages/core/src/components/CheckBox.story.tsx
```diff
@@ -68,12 +68,22 @@ storiesOf('Core/CheckBox', module)
...
+  .add('As a clickable button.', () => (
     <CheckBox
       name="cb-basic"
       label="Label"
       labelDescription="This is a label description."
       onChange={action('onChange')}
       button
     />
+  ))
+  .add('As a compact, clickable button.', () => (
```
stefhatcher: add a story update for RadioButton too?

davidchang:
PTAL @milesj ?

davidchang:
Thanks @stefhatcher - if okay, can you merge ? (i don't have write access)

PR #179, opened 2019-09-06
update(DataTable): add columnHeaderBold prop

gaarf:
to: @milesj @stefhatcher @hayes @alecklandgraf @piewell @schillerk 

## Description

Adds a `columnHeaderBold` prop to make headers bolder and more bordery.

## Motivation and Context

requested by @adamirl 

## Testing

none yet, please advise

## Screenshots

Screen Shot 2019-09-06 at 3 12 22 PM

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/ColumnLabels.tsx
+++ b/packages/core/src/components/DataTable/ColumnLabels.tsx
@@ -29,6 +29,7 @@ type ColumnLabelsProps = {
     we clone them from props (children[0] = label, children[1] = carets), build around their data. */
 export default function ColumnLabels({
   cx,
+  columnHeaderBold,
   styles,
   columnToLabel = {},
   showColumnDividers,
@@ -43,6 +44,7 @@ export default function ColumnLabels({
   styles: WithStylesProps['styles'];
   columnToLabel?: ColumnToLabel;
   showColumnDividers?: boolean;
+  columnHeaderBold?: boolean;
   rowHeight?: RowHeightOptions;
   columnHeaderHeight?: HeightOptions;
   expandable?: boolean;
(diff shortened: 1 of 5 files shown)


packages/core/src/components/DataTable/index.tsx
```diff
@@ -447,6 +448,9 @@ export default withStyles(
       borderBottom: ui.border,
       cursor: 'pointer',
     },
+    column_header_bold: {
+      borderBottomWidth: ui.borderWidthThick,
```
piewell: Personally I think this should be the default, and would help distinguish the table header from the rest of the table rows.

packages/core/src/components/DataTable/ColumnLabels.tsx
```diff
@@ -90,7 +92,7 @@ export default function ColumnLabels({
               className={cx(styles && styles.row)}
             >
               <span>
-                <Text micro muted>
+                <Text micro={!columnHeaderBold} muted={!columnHeaderBold} bold={columnHeaderBold}>
```
piewell: This looks good to me since it will avoid any breaking changes

milesj:
Pretty sure this can already be implemented using the `columnToLabel` prop.

milesj:
@piewell `Can just use <b>`.

We try to avoid adding props for every little thing. At some point it will simply be overwhelming.

stefhatcher:
yeah, don't think we need this PR. @schillerk also mentioned you can use the `columnHeaderRenderers` prop if you want bold column headers.

gaarf:
@stefhatcher @milesj ok, what about the thicker border for the header? is that a breaking visual change or a prop or also something we don't want?

stefhatcher:
i'm a 👎 on the thicker border variation. would prefer to keep data table ui consistent. data table use's the theme ui.border now and borderThick is used sparingly. if the folks on data want the thicker border to be the default, that sounds good, but wouldn't want to introduce a new prop for such a minor visual change.

PR #180, opened 2019-09-11
fix(Dropdown): Remove prop type restrictions.

milesj:

## Description

Removes the prop type restrictions. Not sure why these were here since its valid to define them in parallel.

## Motivation and Context

I have a use case where I need to define left and right at the same time, yet it currently errors.

## Checklist

### DIFF
--- a/packages/core/src/components/Dropdown/index.tsx
+++ b/packages/core/src/components/Dropdown/index.tsx
@@ -1,15 +1,6 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import { mutuallyExclusiveProps } from 'airbnb-prop-types';
+import { Block } from 'aesthetic';
 import withStyles, { WithStylesProps } from '../../composers/withStyles';
-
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Dropdown/index.tsx
```diff
@@ -105,7 +89,7 @@ class Dropdown extends React.PureComponent<Props & WithStylesProps> {
       onClickOutside,
       ...props
     } = this.props;
-    const style: React.CSSProperties = {
```
stefhatcher: why wasn't this using withStyles before?

milesj: It was, just the type was wrong.

stefhatcher: aah i expanded the rest of the diff and i see it now.

stefhatcher: 😎

PR #195, opened 2019-09-24
update(Row, CardContent): Add `minHeight` support. Add `small` image option. Rework CardContent before/after images and increase default size.

Alphy11:

## Description

- Changing default size for Content before/after images to be 152px
- Adding small size for Content before/after images at previous default of 80px
- Changing the Image to use background so that the card scales with the content of the card, not the image.
- Add a minHeight to Content

## Motivation and Context

The focus of a card is its' content, and we should scale with that rather than the image.  Also, design decided that we should use 152px as the width.

## Testing

Storybook testing.

## Screenshots
before:
image

after:
image

## Checklist


packages/core/src/components/Row/index.tsx
```diff
@@ -50,6 +52,7 @@ export class Row extends React.Component<Props & WithStylesProps> {
       compact,
       inline,
       maxHeight,
+      minHeight,
```
stefhatcher: mind adding a story to show this new prop too?

packages/core/src/components/Card/Content.tsx
```diff
@@ -24,6 +25,19 @@ const beforePropType = mutuallyExclusiveProps(
   'beforeImageSrc',
 );
 
+const imageSizePropType = mutuallyExclusiveProps(PropTypes.bool, 'small', 'large');
+
+function getSideImageWidth({ small, large }: { large?: boolean; small?: boolean }) {
+  if (small) {
+    return 80;
+  }
+  if (large) {
```
stefhatcher: 🎩 super nit, but add a new line between the if blocks for readability

packages/core/test/components/Card/Content.test.tsx
```diff
@@ -16,22 +16,14 @@ describe('<Card />', () => {
     const imageUrl = 'LeftFoo.jpg';
     const wrapper = shallowWithStyles(<Content beforeImageSrc={imageUrl}>Sup</Content>);
 
-    expect(
-      shallow(wrapper.find(Row).prop('before') as any)
-        .find('img')
-        .prop('src'),
-    ).toBe(imageUrl);
+    expect((wrapper.find(Row).prop('before')! as any).props.src).toBe(imageUrl);
```
stefhatcher: does this work? so we dont need the `! as any`

`expect(shallow(wrapper.find(Row).prop('before')).prop('src')).toBe(imageUrl);`

looks like it had a shallow before

packages/core/src/components/Row/index.tsx
Alphy11: @stefhatcher It seemed like the paradigm was not to add stories for these types of props.  There was no maxHeight story either

Alphy11: Oh I was looking at row.  Sure will add for both

packages/core/test/components/Card/Content.test.tsx
Alphy11: @stefhatcher before we shallowed and asserted on the nested item, since it was wrapped in a div, this was possible.  But now Image is the top level item in the prop, and enzyme does not support asserting props on the wrapper itself.  So I assert on the passed in component.

stefhatcher: gotcha 👍

Alphy11:
@stefhatcher Looks like just happo failing now.

stefhatcher:
@Alphy11 also, looks like there's a conflict with `packages/core/test/components/Card/Content.test.tsx`

Alphy11:
@stefhatcher Okay now forrealsies only happo :-)

stefhatcher:
ah! the happo diffs are helping! you'll need to add a delay, now that the images are using `<Image />`.

here's an example of adding a delay to the stories: https://github.com/airbnb/lunar/blob/master/packages/core/src/components/Image.story.tsx#L9

also curious why the topImage is behaving differently, since that wasn't changed.

packages/core/src/components/Card/Content.tsx
```diff
@@ -215,21 +243,6 @@ export default withStyles(({ color, pattern, ui, unit }) => ({
     paddingLeft: unit * 1.5,
   },
 
-  image: {
```
stefhatcher: top image uses these `image` styles. please bring them back! that's why the top image happo diffs were off

Alphy11:
@stefhatcher good catch, I don't know why that wasn't reflected locally.  These happo tests are pretty slick.  Though, any idea why when I add a delay of even 2 seconds the image still doesn't appear?

packages/core/src/components/Card/Content.tsx
Alphy11: Huh, so weird that this didn't affect my local styleguide at all.

stefhatcher: removing it squished the image instead of covering the space. would be hard to notice unless you have that image memorized! ha

stefhatcher:
it might be an issue with background images. the `Image` component's background image stories are blank too 😬

packages/core/src/components/Card/Content.tsx
Alphy11: Locally it actually looked the same as the after in happo, that's the weird thing.

Alphy11:
@stefhatcher How do you want to handle these happo stories then?

packages/core/src/components/Card/Content.tsx
stefhatcher: interesting. to debug the happo stuff, i cloned your repo and pulled this branch. locally i'm seeing the broken top image (which is the same as the "after" in the happo):

Screen Shot 2019-09-27 at 1 25 56 PM

since it's missing the `objectFit: 'cover'`, it doesn't look as intended:
http://airbnb.io/lunar/?path=/story/core-card--a-card-with-a-top-featured-image

Alphy11:
I definitely confused that the image should look like with what it did in the after :-) whoooops!

packages/core/src/components/Card.story.tsx
```diff
@@ -5,14 +5,17 @@ import IconChevronRight from '@airbnb/lunar-icons/lib/interface/IconChevronRight
 import LoremIpsum from ':storybook/components/LoremIpsum';
 import lunar from ':storybook/images/lunar-logo.png';
 import stars from ':storybook/images/stars.jpg';
-import moon from ':storybook/images/moon.png';
 import Text from './Text';
 import ResponsiveImage from './ResponsiveImage';
 import Card, { Content } from './Card';
 
+const moon =
```
stefhatcher: it was worth a shot to try this :( thank you!

let's remove it/ go back to `import moon from ':storybook/images/moon.png';`

and call this good to go!

packages/core/src/components/Card.story.tsx
```diff
@@ -5,14 +5,15 @@ import IconChevronRight from '@airbnb/lunar-icons/lib/interface/IconChevronRight
 import LoremIpsum from ':storybook/components/LoremIpsum';
 import lunar from ':storybook/images/lunar-logo.png';
 import stars from ':storybook/images/stars.jpg';
-import moon from ':storybook/images/moon.png';
+import moon from ':storybook/images/moon.jpg';
```
stefhatcher: oops - think this is supposed to be `png` not `jpg`

Alphy11:
@stefhatcher I assume the last failing happo in travis is known?

Alphy11:
@stefhatcher Friendly monday morning reminder :-)

stefhatcher:
since happo is broken for forked repo PRs, closing in favor of https://github.com/airbnb/lunar/pull/201

PR #201, opened 2019-09-30
update(Row, CardContent): Add `minHeight` support. Add `small` image option. Rework CardContent before/after images and increase default size.

stefhatcher:
to: @Alphy11 

## Description

Remake of https://github.com/airbnb/lunar/pull/195/ since Happo is broken.

> - Changing default size for Content before/after images to be 152px
> - Adding small size for Content before/after images at previous default of 80px
> - Changing the Image to use background so that the card scales with the content of the card, not the image.
> - Add a minHeight to Content
> - Add a minHeight to Row

## Motivation and Context

> The focus of a card is its' content, and we should scale with that rather than the image.  Also, design decided that we should use 152px as the width.

## Screenshots

before:
image

after:
image

## Checklist

### DIFF
--- a/packages/core/src/components/Image/index.tsx
+++ b/packages/core/src/components/Image/index.tsx
@@ -40,7 +40,7 @@ const styleSheet: StyleSheet = ({ color, ui }) => ({
   fadeIn: {
     animationName: {
       from: {
-        opacity: 0,
+        opacity: 0.1,
       },
 
       to: {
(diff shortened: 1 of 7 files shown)


stefhatcher:
also, @milesj

packages/core/src/components/Image/index.tsx
```diff
@@ -38,7 +40,7 @@ const styleSheet: StyleSheet = ({ ui }) => ({
   fadeIn: {
     animationName: {
       from: {
-        opacity: 0,
```
stefhatcher: small bump allows us to see a little bit of the image in happo, which is helpful for the alignment and sizing. even with a large happo delay, this fadeIn animation doesn't complete.

PR #204, opened 2019-10-07
fix(FormActions): Fix FormActions error `Cannot add property disabled, object is not extensible`.

stefhatcher:

## Description

Tried using FormActions exported from the form package within a Form. Got `Cannot add property disabled, object is not extensible`. Error. Cannot use `Object.assign` on `props`. Updated the story to use the FormActions from the form package, not the core version.

## Testing

locally and in storybook

## Screenshots

imported the FormActions from the form package into the form story to reproduce the error:

Storybook 2019-10-07 16-26-13

## Checklist

### DIFF
--- a/packages/forms/src/components/FormActions/index.tsx
+++ b/packages/forms/src/components/FormActions/index.tsx
@@ -9,16 +9,13 @@ export type Props = Omit<BaseProps, 'disabled' | 'processing'>;
 /** `FormActions` automatically connected to the parent `Form`. */
 export default function FormActions(props: Props) {
   const context = useContext(FormContext);
-  let baseProps: BaseProps = props;
+  const baseProps: BaseProps = { ...props };
 
   if (context) {
     const { submitting, valid }: FormState<unknown> = context.getState();
 
-    baseProps = {
-      ...props,
-      disabled: !valid,
-      processing: submitting,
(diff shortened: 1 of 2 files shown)


packages/forms/src/components/FormActions/index.tsx
```diff
@@ -9,15 +9,17 @@ export type Props = Omit<BaseProps, 'disabled' | 'processing'>;
 /** `FormActions` automatically connected to the parent `Form`. */
 export default function FormActions(props: Props) {
   const context = useContext(FormContext);
+  let baseProps: BaseProps = props;
```
milesj: If you spread here then the assign code doesn't change much.

milesj:
How is this not breaking all over our apps? 🤔

stefhatcher:
it was when i tried to use it in nova and when i used it in storybook (instead of the one from core)

packages/forms/src/components/FormActions/index.tsx
stefhatcher: will update

stefhatcher:
@milesj - found the reason we haven't seen it the break yet? whenever a form element has `unregisterOnUnmount` and FormActions are used from the form package

PR #206, opened 2019-10-08
update(Apollo): Add typeDefs to Apollo settings

piewell:

## Description

Add typeDefs to Apollo settings so we can define the types for local data.
typeDefs example: https://www.apollographql.com/docs/react/data/local-state/#client-side-schema

## Motivation and Context

See above!

## Testing

Suggestions welcome

## Checklist

### DIFF
--- a/packages/apollo/src/index.ts
+++ b/packages/apollo/src/index.ts
@@ -1,10 +1,9 @@
 import Core from '@airbnb/lunar';
-import { ApolloClient, Resolvers } from 'apollo-client';
+import { ApolloClient, ApolloClientOptions } from 'apollo-client';
 import { InMemoryCache } from 'apollo-cache-inmemory';
 import { ApolloLink } from 'apollo-link';
 import { onError } from 'apollo-link-error';
 import { HttpLink } from 'apollo-link-http';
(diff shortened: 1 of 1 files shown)


packages/apollo/src/index.ts
```diff
@@ -16,13 +17,15 @@ export { onError, HttpLink, Mutation, Query, Provider };
 
 export type Settings = {
   links?: ApolloLink[];
-  resolvers?: Resolvers;
+  resolvers?: Resolvers | Resolvers[];
+  typeDefs?: string | string[] | DocumentNode | DocumentNode[];
```
milesj: `typeDefs?: ApolloClientOptions['typeDefs']`

piewell: `ApolloClientOptions` requires 1 type argument. Looking at Apollo, it's `ApolloClientOptions<TCacheShape>`, but I'm not sure where to import `TCacheShape` from. 🤔

piewell: @miles Updated! I'm passing <null> as the value for `TCacheShape`, since we're not using cache here and the parameter is required.

PR #207, opened 2019-10-09
deps: Audit 2019-10-09

milesj:

## Description

- Updates all dependencies.
- Moved all `@types` versions in each package to use `*`, which is more accurate and easier to integrate for consumers. The fixed type is now in the root for development.
- Removed `utility-types` since `Omit` is now standard in TS v3.5.

## Motivation and Context

Fall cleaning.

## Checklist


packages/core/src/components/private/FormInput.tsx
```diff
@@ -21,10 +20,11 @@ export type IgnoreAttributes =
   | 'typeof'
   | 'vocab'
   // Non-standard attributes
+  | 'color'
+  | 'css'
```
milesj: Emotions `css` prop (from Storybook) is breaking our declaration files.

stefhatcher: 🤦‍♂

PR #208, opened 2019-10-10
fix: Rewrite stories to remove `@emotion/core` unintended import.

milesj:
to: @milesj @stefhatcher @hayes @alecklandgraf

## Description

Removes `@storybook` imports from all TS stories, and rewrites the Lightbox story to the new format.

## Motivation and Context

Our `.d.ts` files were including a `/// <reference types="@emotion/core" />` for some odd reason, which would break consumers build. Turns out Storybook is the cause of this, so we had to remove all Storybook references within TS files.

## Checklist

- My code follows the style guide of this project.
- I have updated or added documentation accordingly.
- I have read the CONTRIBUTING document.

PR #210, opened 2019-10-10
fix: Rewrite stories to remove `@emotion/core` unintended import.

milesj:

## Description

Happo broken here: https://github.com/airbnb/lunar/pull/208

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,12 +28,12 @@ jobs:
         with:
           node-version: 12
       - uses: ./.github/actions/setup
-      - name: Using current SHA because no previous SHA
-        if: contains(github.event.before, '0000000000') == true
-        run: echo ::set-env name=PREVIOUS_SHA::"${{ github.event.after }}"
       - name: Using previous SHA
         if: contains(github.event.before, '0000000000') == false
         run: echo ::set-env name=PREVIOUS_SHA::"${{ github.event.before }}"
+      - name: Using master SHA because no previous SHA
+        if: contains(github.event.before, '0000000000') == true
+        run: echo ::set-env name=PREVIOUS_SHA::$(git merge-base origin/master HEAD)
       - name: Run Happo CI
         run: yarn run happo:ci
         env:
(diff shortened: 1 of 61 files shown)


trotzig:
Hi there 👋!

I'm looking into the happo issues right now. I hope to have an update soon.

trotzig:
Okay I think I know what the problem is now. It looks like it's actually a problem in the storybook setup and not on the happo side of things. Although it's confusing that happo is showing all examples as deleted (I'll file an internal issue about fixing that on our end). 

This is what the storybook page looks like in the `fix-dts-3` branch:
Screen Shot 2019-10-11 at 11 18 12

In 2bf9bb2, the `action` import is removed from all stories, and instead set as a global. In `packages/core/src/components/MenuToggle/story.tsx` however (on [this line](https://github.com/airbnb/lunar/blob/6985c6ad74c521d062f9cfd32e0a1395f82a0481/packages/core/src/components/MenuToggle/story.tsx#L6)), we're referencing the `action` global before it's defined. 

I think you can fix this by moving the global up to the top in `.storybook/config.js`: 

```
diff --git a/.storybook/config.js b/.storybook/config.js
index c3c610b..0b43416 100644
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -8,6 +8,8 @@ import { withProps } from './addons/props';
 import { withStory } from './addons/story';
 import contexts from './contexts';
 
+global.action = key => (...args) => console.log(key, ...args);
+
 setDefaultDelay(100);
 
 addDecorator(story => <div style={{ padding: 24 }}>{story()}</div>);
@@ -18,4 +20,3 @@ addDecorator(withContexts(contexts));
 
 configure(require.context('../packages', true, /\/([A-Za-z0-9]+\.)?story\.tsx?$/), module);
 
-global.action = key => (...args) => console.log(key, ...args);

```

I hope this helps! I'll look into catching this issue in the happo client.

stefhatcher:
@milesj i pushed up the change to see if it'd fix things! def fixed storybook locally

PR #216, opened 2019-10-22
fix(Dropdown): Omit styles from props before passing into cx

Alphy11:

## Description

withStyles passes in styles, and we do not omit it before passing props into the cx function.  For some reason this breaks dropdown on styleguide, because of some weirdness around them editing function prototypes.  

## Motivation and Context

Styleguide has some library that adds something weird to function prototypes, that cause circular reference to themselves, and the prefix function in aphrodite breaks on a circular reference.  Fixing this is the easiest way to fix that.  And it is a bug we are passing it in anyways.

## Testing

CI

## Checklist

### DIFF
--- a/packages/core/src/components/Dropdown/index.tsx
+++ b/packages/core/src/components/Dropdown/index.tsx
@@ -87,6 +87,7 @@ class Dropdown extends React.PureComponent<Props & WithStylesProps> {
       zIndex,
       visible,
       onClickOutside,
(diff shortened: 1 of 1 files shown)


Alphy11:
I couldn't get lint working locally, sorry if CI fails :-(

stefhatcher:
i can't reproduce the error, but agreed it should be omitted.

PR #220, opened 2019-10-23
update(DataTable): Flip expanded row caret direction.

schillerk:

## Description
Previously carets faced right by default and down when rows are expanded. Eli tells me this is actually flipped and should be the opposite.

Here's the new orientation:
Screen Shot 2019-10-22 at 10 11 25 PM

## Checklist


stefhatcher:
why are the happos diffs showing this change on accordion too??

schillerk:
Thanks so much for catching that, I should have noticed it when I submitted the change. After consulting with Eli, we're going to pick consistency and abandon this change.

stefhatcher:
no worries! and sounds good 👍

elibrumbaugh:
Thanks for flagging @stefhatcher 💯

PR #224, opened 2019-10-28
update(Table): Add `noWrap` prop.

markshlick:

## Description

Add a flag to disable the responsive wrapper div's overflow rules.

## Motivation and Context

When I was rendering menus in my table, I was seeing overflow and clipping when these styles were enabled.

## Testing

Added a unit test that assets against rendered classnames.

## Screenshots

[See Slack thread]

## Checklist

### DIFF
--- a/packages/core/src/components/Table/index.tsx
+++ b/packages/core/src/components/Table/index.tsx
@@ -20,6 +20,8 @@ export type Props = {
   loading?: boolean;
   /** Applies verticalAlign: middle to child cells. */
   middleAlign?: boolean;
+  /** Disable responsive wrapper. */
+  noWrap?: boolean;
   /** Alternate row background color. */
   striped?: boolean;
   /** A unique name for tracking purposes. */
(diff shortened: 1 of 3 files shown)


milesj:
I don't agree with this implementation or the terminology.

stefhatcher:
how so @milesj ?

milesj:
The wrapper exists for scenarios where the parent or viewport is too small, and the table is to wide, so it wraps it in a scrollable area. By using no wrap, it will basically break the page under those scenarios.

packages/core/src/components/Table/index.tsx
```diff
@@ -20,6 +20,8 @@ export type Props = {
   loading?: boolean;
   /** Applies verticalAlign: middle to child cells. */
   middleAlign?: boolean;
+  /** Disable responsive wrapper. */
+  noWrap?: boolean;
```
stefhatcher: a story with `noWrap` and rows with MenuToggle in it to demonstrate these two working together, would be great!

markshlick: PTAL!

PR #226, opened 2019-10-29
new(DataTable): Add `sortByValue` accessor + `sortByCacheKey`

williaster:
to: @milesj @stefhatcher
cc: @schillerk 

## Motivation and Context

tl;dr **`DataTable` sorting + expanded rows are incompatible** in some cases

Currently `sortBy` values are static in `DataTable`, meaning that you cannot dynamically change the value used to determine the sort order of a cell. We have use cases where the value of a cell changes dynamically (without a `data` update), which should result in a new sort order.

The current options for updating the sort are either 
a) pass new `data` to the table with the column value 
b) handle sort in the parent with `sortOverride`, which also requires new `data`. 

The primary issue we've encountered with new `data` is that is clears expanded + selected rows (this is even more problematic when `dynamicRowHeight` is used). This (again) makes sorting + expanded rows incompatible.

## Description
This PR adds two props to `DataTable` to address the above limitation:
- `sortByValue: (rowData: T, columnKey: string) => unknown` is a new function that returns the value used for sorting rows, it defaults to the previous behavior of `rowData.data[columnKey]`
- `sortByCacheKey` Key used as part of the sorted data cache key, which can be used to force re-sorting without a full `sortOverride` (and the required new `data`).

It also adds a new story to demonstrate.

## Testing

- [x] CI
  - [x] new unit tests (WIP)

- [x] Functional new storybook example
  - [x] `sortByValue` determines sort order
  - [x] `sortByCacheKey` changes 
  - [x] `expandedRow` state is maintained across sorts

## Screenshots

New storybook example
dynamic-sort-key

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/DataTable.tsx
+++ b/packages/core/src/components/DataTable/DataTable.tsx
@@ -120,8 +120,9 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
       sortBy: string,
       sortDirection: SortDirectionType,
       selectedRows: SelectedRows,
+      maybeSortKey?: string,
     ): IndexedParentRow[] => {
-      const { selectedRowsFirst } = this.props;
+      const { selectedRowsFirst, sortByValue } = this.props;
       const indexedData = indexData(data);
       const sortedData = sortData(
         indexedData,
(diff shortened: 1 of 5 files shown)


packages/core/src/components/DataTable/DataTable.tsx
```diff
@@ -120,8 +120,9 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
       sortBy: string,
       sortDirection: SortDirectionType,
       selectedRows: SelectedRows,
+      maybeSortKey?: string,
```
williaster: oops gotta remove this

PR #231, opened 2019-11-05
fix(DataTable): Clear CellMeasureCache on sort change

williaster:
to: @milesj @stefhatcher 
cc: @schillerk 

## Motivation and Context

This PR fixes another issue with dynamic row height + sorting 😭 The cell measure cache should be invalidated on sort change when `dynamicRowHeight` is used, otherwise you can end up with invalid row heights (see images below).

## Description

This just adds an additional check in `componentDidUpdate` to clear the cache if sorts were changed, and updates the story to demonstrate that it works.

## Testing

- [x] CI
- [x] functional in storybook (see screenshots)

## Screenshots

**Before**
(note the extra large row on top + clipped expanded row on bottom when the sort key is changed)
dynamic-sort-key-row-size--before

**After**
dynamic-sort-key-row-size--after

## Checklist


williaster:
the diff is expected due to enabling `dynamicRowHeight`

PR #233, opened 2019-11-07
update: Refactor form components to use hooks

milesj:

## Description

Rewrites all form components to use the new `useFormField` hook instead of the `connectToForm` HOC.

## Motivation and Context

- Hooks are easier to use than HOCs.
- Hooks type better.
- Hooks are easily testable.
- Slightly reduces filesize.
- Generics can now be passed properly (Auto/multicomplete were broken before).

## Testing

Storybook.

## Checklist

### DIFF
--- a/packages/forms/src/components/Form/index.tsx
+++ b/packages/forms/src/components/Form/index.tsx
@@ -356,15 +356,10 @@ export default class Form<Data extends object = {}> extends React.Component<
 
     field.data.config = config;
     // @ts-ignore
-    field.initial = value;
-    // @ts-ignore
     field.value = value;
     field.touched = config.validateDefaultValue || false;
-
-    /* eslint-disable no-param-reassign */
-    formState.initialValues = setIn(formState.initialValues!, name, value);
+    // eslint-disable-next-line no-param-reassign
     formState.values = setIn(formState.values, name, value);
-    /* eslint-enable no-param-reassign */
   }
 
   /**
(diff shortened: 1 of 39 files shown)


milesj:
@stefhatcher @alecklandgraf This is ready.

milesj:
Those stats are wrong ^ I think the build files are different in CI than on my local.

packages/forms/src/components/Form/index.tsx
```diff
@@ -356,15 +356,10 @@ export default class Form<Data extends object = {}> extends React.Component<
 
     field.data.config = config;
     // @ts-ignore
-    field.initial = value;
-    // @ts-ignore
     field.value = value;
     field.touched = config.validateDefaultValue || false;
-
-    /* eslint-disable no-param-reassign */
-    formState.initialValues = setIn(formState.initialValues!, name, value);
```
stefhatcher: this might be a breaking change for apps that use initialValues?

milesj: It still works (look a few lines above). This was just setting the value back to itself.

stefhatcher:
noticed a few differences in the form example from what's currently implemented:
- rest totally clears the form, instead of resetting to default/initial values 
- date time picker's default value isn't working the same, e.g. can fill out all other form elements and the form is considered value, but in this branch, have to change date time picker's default value for the form to be valid (before "save" becomes active)
- in core, DatePickerInput's example of two inputs: doesn't close the first one after selecting a value before opening up the second picker dropdown

packages/forms/src/components/Form/index.tsx
milesj: This actually broke form resets.

PR #236, opened 2019-11-08
feature: Update simple components to use `useStyles` instead of `withStyles`.

stefhatcher:

## Description

- converting non-lifecycle, `withStyles` components to functional ones to use `useStyles` instead
- moved stylesheets into co-located `styles.ts` files

WIP updating tests

## Motivation and Context

perf

## Testing

local storybook, but need to be more thorough
happo

## Checklist


packages/core/src/components/Alert/index.tsx
```diff
@@ -42,173 +43,91 @@ export type Props = {
...
-
-    alert_statusIcon: {
-      paddingLeft: unit * 2.5,
-    },
-  }),
-  {
-    passThemeProp: true,
-  },
-)(Alert);
+Alert.propTypes = {
```
milesj: If you define prop types like this, the `export default` will need to happen after, otherwise the TS declarations are broken.

stefhatcher: oh yeah, will update! just a few of these deviated from the typical conversion

stefhatcher: fixed!

packages/core/src/components/Alert/styles.ts
```diff
@@ -0,0 +1,65 @@
+import { StyleSheet } from '../../hooks/useStyles';
+
+const styleSheet: StyleSheet = ({ color, pattern, unit }) => ({
```
milesj: Maybe we use named exports everywhere? Just for situations where there may be multiple sheets.

packages/core/src/components/ButtonGroup/index.tsx
```diff
@@ -1,35 +1,7 @@
...
-            <div className={cx(stacked ? styles.cell_stacked : styles.cell)}>{child}</div>
-          ) : null,
-        )}
-      </div>
-    );
-  }
-}
-
-export default withStyles(({ unit }) => ({
+const styleSheet: StyleSheet = ({ unit }) => ({
```
milesj: This one inline?

packages/core/src/components/Alert/styles.ts
stefhatcher: i was thinking `styleSheet` for `index` styles. and then for multiple sheets, used the file name as a suffix, e.g. for Item in Accordion: `styleSheetItem`.  thoughts on that?

packages/core/src/components/ButtonGroup/index.tsx
stefhatcher: oops, yeah. this must've been before the switch to `styles.ts`. will update! good eye

packages/core/src/components/Alert/styles.ts
milesj: Yeah sounds. Just think we should align all styles to use named exports.

stefhatcher: @milesj cool - ill work on doing this approach for the others

stefhatcher: done!

stefhatcher:
nice diff 
Screen Shot 2019-11-12 at 3 58 48 PM

PR #239, opened 2019-11-11
ci: Migrate to Travis CI.

milesj:

## Description

Because of our current Airbnb billing plan, we're unable to use GitHub actions moving forward. Updating our CI build to use Travis instead.

I'm also moving away from DangerJS, as it contains security holes because GH tokens need to be public for forks to work. We can simply do anonymous API requests since all this data is public to begin with.

## Motivation and Context

We have to.

## Testing

N/A

## Checklist


milesj:
@stefhatcher @hayes This is ready for review.

PR #248, opened 2019-11-15
update(CardContent): Allow full size content in before and after for Card Content

Alphy11:

## Description
Adding full size props for before/after to remove automatic spacing in the content.

## Motivation and Context

There's been some churn in using before/after card content, first with us needing a new size for the content, and now us needing to make the image a link.  Making the content a bit more flexible so we can customize it as needed.
## Testing

Checking in storybook.

## Screenshots

image
image

## Checklist

### DIFF
--- a/packages/core/src/components/Card/Content.tsx
+++ b/packages/core/src/components/Card/Content.tsx
@@ -6,6 +6,7 @@ import Image from '../Image';
 import Row from '../Row';
 import Spacing from '../Spacing';
 import { styleSheetContent as styleSheet } from './styles';
+import ButtonOrLink from '../private/ButtonOrLink';
 
 function getSideImageWidth({ large, small }: { large?: boolean; small?: boolean }): number {
   if (small) {
@@ -107,9 +108,9 @@ function CardContent({
 
     if (onAfterImageClick) {
       afterContent = (
-        <button name="AfterContent" className={cx(styles.side_button)} onClick={onAfterImageClick}>
+        <ButtonOrLink className={cx(styles.sideButton)} onClick={onAfterImageClick}>
           {afterContent}
-        </button>
+        </ButtonOrLink>
       );
     }
   }
@@ -141,13 +142,9 @@ function CardContent({
 
     if (onBeforeImageClick) {
       beforeContent = (
-        <button
-          name="BeforeContent"
-          className={cx(styles.side_button)}
-          onClick={onBeforeImageClick}
-        >
+        <ButtonOrLink className={cx(styles.sideButton)} onClick={onBeforeImageClick}>
           {beforeContent}
-        </button>
+        </ButtonOrLink>
       );
     }
   }
--- a/packages/core/src/components/Card/Content.tsx
+++ b/packages/core/src/components/Card/Content.tsx
@@ -46,6 +46,10 @@ export type Props = {
   topImageSrc?: string;
   /** To use with text truncation; overflow is hidden. */
   truncated?: boolean;
+  /** If provided, makes the after image clickable, firing this callback. */
+  onAfterImageClick?: () => void;
+  /** If provided, makes the before image clickable, firing this callback. */
+  onBeforeImageClick?: () => void;
   /** If provided, makes the entire content clickable, firing this callback. */
   onClick?: () => void;
 };
@@ -66,6 +70,8 @@ function CardContent({
   small,
   topImageSrc,
   truncated,
+  onAfterImageClick,
+  onBeforeImageClick,
 }: Props) {
   const [styles, cx] = useStyles(styleSheet);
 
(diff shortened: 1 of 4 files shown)


packages/core/src/components/Card/styles.ts
```diff
@@ -49,6 +49,10 @@ export const styleSheetContent: StyleSheet = ({ color, pattern, ui, unit }) => (
     paddingBottom: unit * 1.5,
   },
 
+  side_full_size: {
```
stefhatcher: please update to `side_fullSize` - the `_` is used to signal a modifier, then camelCase naming continues

stefhatcher:
i think there should be support to add an on click to the beforeImgSrc/afterImgSrc, not open up before/after to be full sized. this'll help keep images in cards consistent, but still add onClick support.

packages/core/src/components/Card/story.tsx
```diff
@@ -332,3 +332,61 @@ export function cardAsAButtonWithMiddleAlignment() {
...
+};
+
+export function asFullSizeCustomAfter() {
+  return (
+    <Card>
+      <Content
+        truncated
+        afterFullSize
+        after={
+          <ResponsiveImage
```
stefhatcher: this and the above examples of `afterFullSize` + `after` is doing what `afterImageSrc` does. if the goal is to make the before/after images clickable, add `onClickBeforeImage` and `onClickAfterImage` to keep before/after image treatments consistent. before/after content is supposed to have the padding.

Alphy11: It seems like enforcing style here might not be a scalable approach.  When we could make the component more flexible and less complex by removing before/after image src, as well as the large/small props, and just allowing you to define your own padding.  I can certainly just start adding more props on to do what we want, but it feels like it will just continue to bloat the component

@stefhatcher

stefhatcher: yeah, but then we'd lose consistency with images in cards in the design system.

milesj: We should just redo card for v3. It's become to unmanageable.

Alphy11: I would agree with that.  Short term both of you agree I should go and add click handler?

Alphy11: If so, any recommendation on defining the prop type there?  I am not super familiar with airbnb prop types library, and it feels complex since it's mutually exlcusive with before/after, but also requires beforeImageSrc/afterImageSrc

stefhatcher: since we'll do a re-work, maybe it's fine to just do the check `onClickBeforeImageSrc && beforeImageSrc` in the code, no extra prop type def besides the callbacks being functions

Alphy11:
Is there an established pattern to remove the post-click dropshadow on a button?  I thought the button reset pattern was supposed to, but I am still seeing that purple dropshadow when I click the image.

packages/core/src/components/Card/Content.tsx
```diff
@@ -98,6 +104,14 @@ function CardContent({
         src={afterImageSrc}
       />
     );
+
+    if (onAfterImageClick) {
+      afterContent = (
+        <button name="AfterContent" className={cx(styles.side_button)} onClick={onAfterImageClick}>
```
stefhatcher: curious why the `name` attributes?

packages/core/src/components/Card/styles.ts
```diff
@@ -44,6 +44,11 @@ export const styleSheetContent: StyleSheet = ({ color, pattern, ui, unit }) => (
     paddingBottom: unit * 3,
   },
 
+  side_button: {
```
stefhatcher: `side_button` -> `sideButton` please, it is its own element

packages/core/src/components/Card/Content.tsx
stefhatcher: `type="button"` instead?

stefhatcher:
if you use `ButtonOrLink` instead of the `<button />` with styles, it'll remove the focus after click:

https://github.com/airbnb/lunar/blob/master/packages/core/src/components/private/ButtonOrLink.tsx#L2

packages/core/src/components/Card/Content.tsx
Alphy11: heartbeat

milesj: I don't agree with coupling internal technologies to an open source project.

stefhatcher: agreed

Alphy11: Then please make sure in the card rewrite we have the ability to add whatever content we want to the side, because we are going to want to track the clicks on this.

Alphy11:
@milesj @stefhatcher Would you like us to accept our own happo reviews, or would you prefer we leave that to you guys?

Alphy11:
Sorry wrong PR, but question is still valid

PR #258, opened 2019-11-26
new: Add document manager for easier mutation updates.

milesj:

## Description

- Standardizes our updater types.
- Adds a document manager for making updates easier to use.

## Motivation and Context

Mutations support an [update option](https://www.apollographql.com/docs/react/data/mutations/#updating-the-cache-after-a-mutation), which allows the internal cache to be modified. We make use of this with the `addToList` and `removeFromList` updaters in Lunar.

However, for this to work correctly, a query + *exact* variables need to be used, otherwise the query cannot be found in the cache. This can be extremely complicated to achieve, especially if the query and mutation occur in completely different parts of the tree. To mitigate this, the document manager can be used to cache queries by a unique key, in which the updater can use.

```ts
update: addToList('some-query-on-another-page', 'data.list', 'saveData'),
```

## Testing

Unit tests.

## Checklist

### DIFF
--- a/packages/apollo/src/DocumentManager.ts
+++ b/packages/apollo/src/DocumentManager.ts
@@ -1,0 +1,24 @@
+import { DocumentNode } from 'graphql';
+
+interface QueryCache<T> {
+  query: DocumentNode;
+  variables?: T;
+}
+
+export default class DocumentManager {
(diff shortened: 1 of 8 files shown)


milesj:
Dont need this at this time.

PR #261, opened 2019-12-06
update(Tooltip): Reduce offscreen pattern layer size and render a11y content in Portal

williaster:
to: @milesj @stefhatcher @lencioni @etr2460

## Motivation and Context

We've been trying to do some perf optimization in a data-rich app, and noticed that the `left: -9999` style from `pattern.offset` used by `Tooltip` creates a very large layer which can become expensive for transitions and memory usage.

image

## Description

This PR makes a couple updates to help fix this
- update `pattern.offset` to be `left: -5vw` instead of `-9999px` to reduce its size
- render the persistent offscreen `a11y` tooltip content inside a `<Portal />` so that no matter its size it doesn't affect the flow of the content `<Tooltip />` wraps / is applied to 
- updates a `ref.getBoundingClientRect()` call to be within a `requestAnimationFrame` callback because it for [forces layout/reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#box-metrics) (one call in `Tooltip` was in a `raf` callback, but the other wasn't)

## Testing
- [x] functional 
  - [x] tooltips still function and are correctly positioned
  - [x] `a11y` content is still rendered (now in portal) 
- [x] CI
- [ ] perf
  - I **wasn't** actually able to re-produce the large `Tooltip` layer size in storybook examples unless I set `left: 9999` (instead of `-9999`), but there are no large layers with the portal + `pattern.offscreen` changes. I'm not sure if there is a storybook style that is preventing this 🤔 
  - In storybook it was also tough to test the impact of the new `requestAnimationFrame` on mouseover. I tried profiling but didn't see a significant difference even when rendering ~50 tooltips. If you have any thoughts let me know!

## Checklist

### DIFF
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -156,7 +156,7 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
       const { current } = this.containerRef;
 
       /* istanbul ignore if: refs are hard */
-      if (current) {
+      if (current && this.mounted) {
         this.setState({ targetRect: current.getBoundingClientRect() });
       }
     });
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -6,6 +6,7 @@ import NotchedBox, { NOTCH_SIZE, NOTCH_SPACING } from '../NotchedBox';
 import Text from '../Text';
 import withStyles, { WithStylesProps } from '../../composers/withStyles';
 import { styleSheet } from './styles';
+import Portal from '../Portal';
 
 const EMPTY_TARGET_RECT: ClientRect = {
   bottom: 0,
@@ -76,7 +77,7 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
 
   mounted: boolean = false;
 
-  rafHandle?: number;
+  rafHandle: number = 0;
 
   static getDerivedStateFromProps({ disabled }: Props) {
     if (disabled) {
@@ -105,7 +106,7 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
 
   componentWillUnmount() {
     this.mounted = false;
-    cancelAnimationFrame(this.rafHandle as number);
+    cancelAnimationFrame(this.rafHandle);
   }
 
   updateTooltipHeight() {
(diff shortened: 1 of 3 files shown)


packages/core/src/themes/buildTheme.ts
```diff
@@ -95,7 +95,7 @@ export default function buildTheme(
       },
       offscreen: {
         position: 'absolute',
-        left: -9999,
+        left: '-5vw',
```
stefhatcher: curious why `5`? instead of `50vw` or something a little larger?

williaster: `5` is arbitrary to some extent, but mostly trying to minimize size. `offscreen` specifies a `width`/`height` of `1px` and `overflow: hidden`, so I'd guess `5vw` is okay. 

what do you think of `-10vw` to be safe but smaller than `-50vw`?

stefhatcher: good call! `5` sounds good then!

williaster:
I can't really see any diff changes even tho they were flagged, I don't know the happo algorithm but perhaps it's picking up on the new portal `div`s or the `left` offset change.

milesj:
For the content div, I wonder if we gave it a height and width of like 1, would it help with perf issues.

williaster:
@milesj it should [already have](https://github.com/airbnb/lunar/pull/261/files#diff-9712e6ef9c78106ea51ab21b6c40df70R99) those dimensions

milesj:
Weird... wonder why that would cause issues then.

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -150,12 +151,15 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
...
+    /* istanbul ignore next: refs are hard */
+    this.rafHandle = requestAnimationFrame(() => {
+      const { current } = this.containerRef;
 
-    /* istanbul ignore if: refs are hard */
-    if (current) {
-      this.setState({ targetRect: current.getBoundingClientRect() });
-    }
+      /* istanbul ignore if: refs are hard */
+      if (current) {
```
milesj: Probably need `current && this.mounted`.

williaster: we cancel this callback when we set `this.mounted=false`, but probably worth it to add. updated.

williaster:
@milesj I think the perf concern comes from the creation of such a large layer due to the a11y `content` _offset_ not its size. I'm not sure why this can result in such a large layer given that its `position` is `absolute`, maybe @lencioni has more specific thoughts on this. 

But that's the hope with reducing the offset and also rendering it in a portal so it doesn't affect the flow of the react tree it's rendered in.

packages/core/src/themes/buildTheme.ts
lencioni: @williaster Instead of using a negative left, I think it would be better to use a positive right > 100. That way the width of the element won't cause it to sometimes be within the viewport. You'll want to add a few vw to account for things that bleed, like box shadow. I'd go with:

```suggestion
        right: '105vw',
```

lencioni:
@williaster is correct. The layer is created by the relative container of this element, and because it was positioned so far outside the viewport, the layer had to be large enough to include it. I would wager that rendering it inside a portal solves this completely. The reduced positioning is also a good idea. Thanks for fixing!

PR #262, opened 2019-12-06
new(DataTable): Expose `overscanRowCount` prop

williaster:
to: @milesj @stefhatcher @conglei @schillerk 

## Description

This simply exposes the `react-virtualized` `overscanRowCount` prop in the `DataTable` component.

## Motivation and Context

We have a case where we'd like to change the `overscanRowCount` prop for performance optimization, but currently this prop is not exposed.

## Testing

- [x] CI
  - [x] new jest unit test

## Screenshots

N/A

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/DataTable.tsx
+++ b/packages/core/src/components/DataTable/DataTable.tsx
@@ -57,6 +57,7 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
     instantEdit: true,
     keys: [],
     minimumDynamicRowHeight: undefined,
+    overscanRowCount: 2,
     renderers: {},
     rowHeight: 'regular',
     selectable: false,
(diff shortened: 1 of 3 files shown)


packages/core/src/components/DataTable/DataTable.tsx
```diff
@@ -452,21 +453,22 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
   render() {
     const {
       autoHeight,
+      cx,
```
williaster: just `abc`-ed these

packages/core/src/components/DataTable/types.ts
```diff
@@ -58,6 +58,8 @@ export interface DataTableProps {
   columnMetadata?: ColumnMetadata;
   /** Array of data rows. */
   data?: ParentRow[];
+  /** Ref to the underlying DataTable instance. */
+  dataTableRef?: DataTableRef;
```
williaster: ditto `abc`-ed

PR #270, opened 2019-12-19
fix(Tooltip, Overlay): Fix reflows on cDM, cDU, and render

alecklandgraf:
cc: @gaarf @lencioni @williaster 

## Description

Profiling found forced reflows from Tooltip and Overlay. It's hard to say how this affects load, but profiling this causes more layout renders than necessary during page load. My guess is in the order of 0.5 to 1 seconds could be saved based on reported render times. 

Here's a list of what will cause reflows: https://gist.github.com/paulirish/5d52fb081b3570c81e3a

image
**With a 6x CPU throttle about 10% of the page load is rendering.** 

image
**Here's what one reflow looks like from Tooltip**

## Motivation and Context

Defers calculating page layout via a requestAnimationFrame or until the layout calc is required for rendering.

## Testing

I will update tests after I reproduce this locally in lunar local dev. 

## Screenshots

N/A

## Checklist

### DIFF
--- a/packages/core/src/components/Overlay/index.tsx
+++ b/packages/core/src/components/Overlay/index.tsx
@@ -46,12 +46,11 @@ export default class Overlay extends React.PureComponent<Props, State> {
       this.rafHandle = requestAnimationFrame(() => {
         // getBoundingClientRect casues a reflow
         const { x, y } = current.getBoundingClientRect() as DOMRect;
-        if (x !== this.state.x) {
-          this.setState({ x });
-        }
 
-        if (y !== this.state.y) {
-          this.setState({ y });
+        if (x !== this.state.x || y !== this.state.y) {
+          this.rafHandle = requestAnimationFrame(() => {
+            this.setState({ x, y });
+          });
         }
       });
     }
--- a/packages/core/src/components/Overlay/index.tsx
+++ b/packages/core/src/components/Overlay/index.tsx
@@ -48,9 +48,8 @@ export default class Overlay extends React.PureComponent<Props, State> {
         const { x, y } = current.getBoundingClientRect() as DOMRect;
 
         if (x !== this.state.x || y !== this.state.y) {
-          this.rafHandle = requestAnimationFrame(() => {
-            this.setState({ x, y });
-          });
+          // wrapping this in a second rAF caused the tooltip to render at 0,0 for a frame
+          this.setState({ x, y });
         }
       });
     }
--- a/packages/core/src/components/Overlay/index.tsx
+++ b/packages/core/src/components/Overlay/index.tsx
@@ -32,6 +32,8 @@ export default class Overlay extends React.PureComponent<Props, State> {
     y: 0,
   };
 
+  rafHandle: number = 0;
+
   ref = React.createRef<HTMLDivElement>();
 
   scrollers: ArrayOfScrollables = [];
@@ -41,15 +43,17 @@ export default class Overlay extends React.PureComponent<Props, State> {
 
     /* istanbul ignore next: refs are hard */
     if (current) {
-      const { x, y } = current.getBoundingClientRect() as DOMRect;
+      this.rafHandle = requestAnimationFrame(() => {
+        // getBoundingClientRect casues a reflow
+        const { x, y } = current.getBoundingClientRect() as DOMRect;
+        if (x !== this.state.x) {
+          this.setState({ x });
+        }
 
-      if (x !== this.state.x) {
-        this.setState({ x });
-      }
-
-      if (y !== this.state.y) {
-        this.setState({ y });
-      }
+        if (y !== this.state.y) {
+          this.setState({ y });
+        }
+      });
     }
 
     this.removeScrollListeners();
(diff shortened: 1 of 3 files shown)


alecklandgraf:
Tooltips still work!
image

packages/core/src/components/Overlay/index.tsx
```diff
@@ -41,15 +43,17 @@ export default class Overlay extends React.PureComponent<Props, State> {
...
+      this.rafHandle = requestAnimationFrame(() => {
+        // getBoundingClientRect casues a reflow
+        const { x, y } = current.getBoundingClientRect() as DOMRect;
+        if (x !== this.state.x) {
+          this.setState({ x });
+        }
+
+        if (y !== this.state.y) {
+          this.setState({ y });
+        }
```
lencioni: I think it would be a good idea to put these setStates in another raf. This would probably do the trick:

```js
if (x !== this.state.x || y !== this.state.y) {
  this.rafHandle = requestAnimationFrame(() => {
    this.setState({ x, y });
  });
}
```

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -92,9 +92,13 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
   componentDidMount() {
     this.mounted = true;
 
-    // eslint-disable-next-line react/no-did-mount-set-state
-    this.setState({
-      targetRect: document.body.getBoundingClientRect(),
+    requestAnimationFrame(() => {
```
lencioni:
```suggestion
    this.rafHandle = requestAnimationFrame(() => {
```

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -92,9 +92,13 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
...
+      const targetRect = document.body.getBoundingClientRect();
+
+      // use a second rAF in case setState becomes expensive
```
lencioni: My worry here is more about setState causing layout thrashing

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -190,8 +194,34 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
...
+          </div>
+
+          {/* render off-screen element in a separate layer */}
+          <Portal>
+            <div id={labelID} className={cx(styles.offscreen)}>
+              {content}
+            </div>
+          </Portal>
+        </span>
+      );
```
lencioni: This appears to duplicate some stuff. It would be great to do this in a way that avoids the duplication, to keep the size of the JS smaller, make it easier to maintain, and prevent the two versions from drifting.

alecklandgraf:
I attempted to extract the reference element code into its own component, but it was about the same number of lines in `Tooltip` because of all the props needed, so it didn't feel like a win. 

```diff
@@ -198,22 +199,17 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
     if (!open) {
       return (
         <span ref={this.containerRef} className={cx(styles.container)}>
-          <div
-            aria-labelledby={labelID}
-            className={cx(!disabled && underlined && styles.underlined)}
-            onMouseEnter={this.handleEnter}
-            onMouseLeave={this.handleClose}
+          <ReferenceElement
+            content={content}
+            disabled={disabled}
+            labelID={labelID}
+            underlined={underlined}
+            onClose={this.handleClose}
+            onEnter={this.handleEnter}
             onMouseDown={this.handleMouseDown}
           >
             {children}
-          </div>
-
-          {/* render off-screen element in a separate layer */}
-          <Portal>
-            <div id={labelID} className={cx(styles.offscreen)}>
-              {content}
-            </div>
-          </Portal>
+          </ReferenceElement>
         </span>
       );
     }
```

I'll see if extracting the tooltip/overlay/pop up code works better, the difficulty is making sure all the refs are passed around and typed properly.

i.e. 

```jsx
render() {
  ...
  return (
    ...
    {open && <TooltipComponent {...props} />}
}
```

packages/core/src/components/Overlay/index.tsx
```diff
@@ -46,12 +46,11 @@ export default class Overlay extends React.PureComponent<Props, State> {
       this.rafHandle = requestAnimationFrame(() => {
         // getBoundingClientRect casues a reflow
         const { x, y } = current.getBoundingClientRect() as DOMRect;
-        if (x !== this.state.x) {
-          this.setState({ x });
-        }
 
-        if (y !== this.state.y) {
-          this.setState({ y });
+        if (x !== this.state.x || y !== this.state.y) {
```
alecklandgraf: adding this second rAF here caused a regression where the tooltip is positioned at 0,0 for a frame or two then at its correct position 

image

image

packages/core/src/components/Overlay/index.tsx
alecklandgraf: this caused an issue https://github.com/airbnb/lunar/pull/270#discussion_r360734462

packages/core/src/components/Overlay/index.tsx
alecklandgraf: I think ^^ could be fixed if we didn't default x and y to `0` or tracked when they are set with a flag. 

```jsx
state = { x: null, y: null }
...
// in render
{open && x !== null && y !== null && (
```

PR #274, opened 2020-01-07
fix(Tooltip): Prevent tooltip from rendering at 0, 0

alecklandgraf:
to: @milesj @stefhatcher @hayes @alecklandgraf @gaarf 

## Description

The `Tooltip` can render for one frame at position `0, 0` while `state.targetRect` is being asynchronously set from `componentDidMount` 

## Motivation and Context

Prevent a UI regression.

## Testing

Storybook and existing tests. 

Can log the `state.targetRect` to see its values at render time

**before**
image

## Screenshots

N/A

## Checklist

### DIFF
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -41,6 +41,7 @@ export type State = {
   open: boolean;
   tooltipHeight: number;
   targetRect: ClientRect;
+  targetRectReady: boolean;
 };
 
 export type PositionStruct = {
@@ -69,6 +70,7 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
     open: false,
     tooltipHeight: 0,
     targetRect: EMPTY_TARGET_RECT,
+    targetRectReady: false,
   };
 
   containerRef = React.createRef<HTMLSpanElement>();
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Tooltip/index.tsx
```diff
@@ -185,6 +185,11 @@ export class Tooltip extends React.Component<Props & WithStylesProps, State> {
     const { cx, styles, theme, width: widthProp, content, inverted } = this.props;
     const { open, targetRect, tooltipHeight } = this.state;
 
+    // render null until targetRect is initialized by cDM
+    if (targetRect.top === 0 && targetRect.left === 0) {
+      return null;
+    }
+
```
alecklandgraf: I added this to `renderPopUp` versus in the render at 

```jsx
{open && !(targetRect.top === 0 && targetRect.left ===0) ? this.renderPopUp() : null}
```
because it's more of a concern of the PopUp rendering and less about the business logic knowing it's open. If we extracted the PopUp into its own component, this is where the logic should live.

alecklandgraf:
I have some tests to fix

alecklandgraf:
This is a tiny bit fragile to a tooltip which **should** be at `0, 0`. I could instead use a flag in `this.state` to track this

alecklandgraf:
Still seeing the bug, I think overlay needs patched too

PR #279, opened 2020-01-08
update(HierarchyPicker): Pass the current item object to `onItemPicked`.

milesj:

## Description

Passes the current item/node object as the 2nd argument to `onItemPicked`. The previous 2nd argument is shifted to the 3rd argument (this only exists for legacy usage).

## Motivation and Context

We need a simple way to access the ID or other metadata for the item being selected. Previously, we only passes the path as an array, which required us to manually walk the tree each time. This change will mitigate that.

## Testing

Unit tests and storybook.

## Screenshots

N/A

## Checklist

### DIFF
--- a/packages/core/src/components/HierarchyPicker/story.tsx
+++ b/packages/core/src/components/HierarchyPicker/story.tsx
@@ -285,8 +285,8 @@ class PickerDemo extends React.Component<Partial<Props>, { chosen: Props['chosen
           searchWidth={400}
           searchPlaceholder="Search all the things"
           noResultsLabel="No results match your query."
-          onItemPicked={(nextChosen: string[] | null) => {
-            action('onItemPicked')(nextChosen);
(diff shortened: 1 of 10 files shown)


milesj:
@williaster Since you wrote most of this.

packages/core/src/components/HierarchyPicker/story.tsx
```diff
@@ -285,8 +285,8 @@ class PickerDemo extends React.Component<Partial<Props>, { chosen: Props['chosen
           searchWidth={400}
           searchPlaceholder="Search all the things"
           noResultsLabel="No results match your query."
-          onItemPicked={(nextChosen: string[] | null) => {
-            action('onItemPicked')(nextChosen);
+          onItemPicked={(nextChosen: string[] | null, item: object | null, details?: object) => {
```
williaster: could use `ItemShape` here but `object` could be fine for story

PR #288, opened 2020-01-22
fix(Proofreader): Default prop value to empty string

alecklandgraf:

## Description

Fix a case where `prevProps.value` is `undefined`.

## Motivation and Context

image

## Testing

Added test

## Screenshots

n/a

## Checklist


milesj:
The value comes from the parent `TextArea`. How is that value undefined?

alecklandgraf:
> The value comes from the parent `TextArea`. How is that value undefined?

It can be undefined because the prop `value` is passed in this chain `TextArea -> Proofreader -> BaseTextArea` and while `BaseTextArea` sets a defaultProp for `value`, `TextArea` doesn't.

milesj:
@alecklandgraf Maybe update them all with a default to be safe.

alecklandgraf:
@milesj updated!

PR #291, opened 2020-01-28
new: Polish and improvements.

milesj:

## Description

Will leave comments in the source.

## Motivation and Context

Noticed a few things that could be change while integrating.

## Testing

Yup.

## Checklist

### DIFF
--- a/packages/composer/src/components/Composer.tsx
+++ b/packages/composer/src/components/Composer.tsx
@@ -27,6 +27,10 @@ export type ComposerProps = {
   defaultValues?: Partial<DataSet>;
   /** Whether the input field is disabled or not. */
   disabled?: boolean;
+  /** Placeholder for the email writing mode. */
+  emailPlaceholder?: string;
+  /** Placeholder for the message writing mode. */
+  messagePlaceholder?: string;
   /** Gain a reference to the underlying `textarea`. */
   propagateRef?: InputProps['propagateRef'];
   /** Callback fired when the input value changes. */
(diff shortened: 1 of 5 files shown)


packages/composer/src/components/Composer.tsx
```diff
@@ -27,10 +27,16 @@ export type ComposerProps = {
   defaultValues?: Partial<DataSet>;
   /** Whether the input field is disabled or not. */
   disabled?: boolean;
+  /** Placeholder for the email writing mode. */
+  emailPlaceholder?: string;
+  /** Placeholder for the message writing mode. */
+  messagePlaceholder?: string;
+  /** Gain a reference to the underlying `textarea`. */
+  propagateRef?: InputProps['propagateRef'];
```
milesj: Need a way to access the `textarea`.

packages/composer/src/components/Composer.tsx
```diff
@@ -27,10 +27,16 @@ export type ComposerProps = {
...
   /** Callback fired when the input value changes. */
-  onChange: ChangeHandler;
+  onChange?: ChangeHandler;
   /** Callback fired when the input is submitted. */
-  onSubmit: SubmitHandler;
+  onSubmit?: SubmitHandler;
```
milesj: Both of these are technically not required, so made them optional.

packages/composer/src/components/Input/index.tsx
```diff
@@ -178,6 +190,7 @@ export default function Input({
         />
 
         <Hotkey
+          preventDefault
```
milesj: If we're in a modal or something else that relies on `esc`, then this hotkey would close *everything*. Preventing default so that it only affects the open menu.

PR #305, opened 2020-02-06
update(FilterMenu): Use SecondaryLink for reset.

gaarf:

## Motivation and Context

use `SecondaryLink` instead of `Link muted` for the reset link in `FilterMenu`

## Testing

styleguide

## Screenshots

before / without new prop:
Screen Shot 2020-02-06 at 1 07 49 PM

after :
Screen Shot 2020-02-06 at 3 17 39 PM

## Checklist


packages/forms/src/components/FilterMenu/index.tsx
```diff
@@ -54,6 +56,7 @@ export default function FilterMenu({
   keepOpenOnClear,
   large,
   menuProps,
+  mutedReset = true,
```
stefhatcher: usually boolean value's default is `false`. please rename this to `unmuted`, default of `false` then `muted={!unmuted}`

packages/forms/src/components/FilterMenu/index.tsx
```diff
@@ -141,7 +144,7 @@ export default function FilterMenu({
             <T k="lunar.common.apply" phrase="Apply" context="Apply filters button label" />
           </Link>
 
-          <Link muted type="reset" onClick={handleClear}>
+          <Link muted={mutedReset} type="reset" onClick={handleClear}>
```
stefhatcher: is this related to a design change? it's been a standard pattern to have the clear/reset be muted

gaarf: it's a want for profiles2 / sending you jira in PM

packages/forms/src/components/FilterMenu/index.tsx
```diff
@@ -141,7 +144,7 @@ export default function FilterMenu({
             <T k="lunar.common.apply" phrase="Apply" context="Apply filters button label" />
           </Link>
 
-          <Link muted type="reset" onClick={handleClear}>
+          <Link muted={!unmutedReset} type="reset" onClick={handleClear}>
```
stefhatcher: ah - chatted offline. since the concern is that the muted is too light/looks disabled (agreed), let's replace Link with SecondaryLink, remove muted. this'll make the link darker and match FormActions pattern better (primary color for main action like Submit/Apply, secondary color of Cancel/Clear). no need for a new prop then. better consistency!

PR #315, opened 2020-02-14
new(Accordion): Add secondary styles.

davidchang:

## Description

Adding secondary styles for Accordion.

## Motivation and Context

I had some designs that called for a secondary Accordion (such that it could be nested inside of another Accordion). I'd be happy to show those designs or discuss further.

## Testing

```yarn jest```

## Screenshots

image

## Checklist


milesj:
@davidchang Have these gone through the official design workflows?

Regardless, we rarely add secondary/alternative styles through props, we extend and create a new component.

stefhatcher:
oof, nested accordions? the background color can be added by the app, since accordion no longer applies a background color of its own. curious if it went through the design workflow too? since the title and the content can be customized (smaller text) and there's a no horizontal pattern, is this really needed?

PR #321, opened 2020-02-19
new(Timeline): Add Timeline component

Tgeo:

## Description

Add a new component, `<Timeline />` which shows a chronological set of events.

Open to thoughts on implementation. I decided to go with two different types of "child" items instead of creating one child item component that would have its behavior controlled via boolean props.

Top-level "Timeline" component is empty (just renders `children`) - but seems appropriate for general standardization and for future proofing.

## Motivation and Context

This component already has an implementation internally (Nova) but will be used by other apps, so it makes sense to genericize it for other consumers.

## Testing

Tested locally in storybook.

## Screenshots

image

## Checklist


packages/core/src/components/Timeline/Item.tsx
```diff
@@ -0,0 +1,40 @@
...
+export default function Item({ at, oldest = false, children }: ItemProps) {
+  const theme = useTheme();
+  const [styles, cx] = useStyles(styleSheetItem);
+
+  return (
+    <>
+      <Row middleAlign before={<IconRecord size={8} color={theme.color.core.neutral[3]} />}>
+        <Text small muted>
+          <DateTime medium at={at} />
+          {' ('}
```
milesj: Use `T` and i18n here.

milesj:
@Tgeo We dont support forks. Youll need to get access to the repo and create a native branch.

packages/core/src/components/Timeline/Item.tsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+/** Represents a single item in a collection of chronologically items. */
+export default function Item({ at, oldest = false, children }: ItemProps) {
+  const theme = useTheme();
+  const [styles, cx] = useStyles(styleSheetItem);
+
+  return (
+    <>
+      <Row middleAlign before={<IconRecord size={8} color={theme.color.core.neutral[3]} />}>
```
stefhatcher: add `decorative` prop for the icon

packages/core/src/components/Timeline/SimpleItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const theme = useTheme();
+  const [styles, cx] = useStyles(styleSheetSimpleItem);
+
+  return (
+    <>
+      <Row
+        middleAlign
+        before={
+          <div className={cx(styles.iconWrapper)}>
+            <IconRecord size={4} color={theme.color.core.neutral[3]} />
```
stefhatcher: missing `decorative` as well

packages/core/src/components/Timeline/index.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export type TimelineProps = {
+  /** Collection of timeline items. */
+  children: NonNullable<React.ReactNode>;
+};
+
+/** A chronologically ordered list of items. Ordering is determined by caller. */
+export default function Timeline({ children }: TimelineProps) {
+  return <>{children}</>;
```
stefhatcher: this wrapping tag should be `<ol>` - this is an ordered list of items (by time). the <Item> component should be wrapped with `<li>` tags

packages/core/src/components/Timeline/Item.tsx
```diff
@@ -0,0 +1,40 @@
...
+  children: NonNullable<React.ReactNode>;
+};
+
+/** Represents a single item in a collection of chronologically items. */
+export default function Item({ at, oldest = false, children }: ItemProps) {
+  const theme = useTheme();
+  const [styles, cx] = useStyles(styleSheetItem);
+
+  return (
+    <>
```
stefhatcher: should be an `li`

packages/core/src/components/Timeline/SimpleItem.tsx
```diff
@@ -0,0 +1,36 @@
...
+export type SimpleItemProps = {
+  /** Whether an item is the oldest (chronologically) in the collection of items. */
+  oldest?: boolean;
+
+  /** Content to render for the title of the item. */
+  children: NonNullable<React.ReactNode>;
+};
+
+/** Represents a visually deemphasized single item in a collection of chronologically items. */
+export default function SimpleItem({ oldest = false, children }: SimpleItemProps) {
```
stefhatcher: i think this should be consolidated into `Item` and the difference flagged by a prop called `secondary`

stefhatcher: it's easier to maintain the two together and helps the developer experience (using a flag instead of importing and rendering a different item secondary, deemphasized items). boolean flags for slight visual differences would be more consistent with lunar practices

packages/core/src/components/Timeline/styles.ts
```diff
@@ -0,0 +1,28 @@
+import { StyleSheet } from '../../hooks/useStyles';
+
+export const styleSheetItem: StyleSheet = ({ color, unit }) => ({
+  wrapper: {
+    marginLeft: unit * 0.5 - 1,
+    borderLeft: `2px solid ${color.core.neutral[3]}`,
```
stefhatcher: use this instead: `ui.borderThick `

packages/core/src/components/Timeline/styles.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  wrapper_oldest: {
+    paddingLeft: unit * 3,
+  },
+});
+
+export const styleSheetSimpleItem: StyleSheet = ({ color, unit }) => ({
+  wrapper: {
+    marginLeft: unit * 0.5 - 1,
+    borderLeft: `2px solid ${color.core.neutral[3]}`,
```
stefhatcher: use this instead: `ui.borderThick `

packages/core/src/components/Timeline/styles.ts
```diff
@@ -0,0 +1,28 @@
...
+    paddingLeft: unit * 2.5,
+    height: unit * 2.75,
```
stefhatcher: is this heigh restriction necessary? both `wrapper` styles are pretty similar except for the padding bottom. if consolidated with paddingBottom, is the height still needed?

packages/core/src/components/Timeline/index.tsx
Tgeo: That's a good idea

Tgeo:
https://github.com/airbnb/lunar/pull/321#issuecomment-588133024

@milesj what's the process for getting added to the repo? Unsure if I am already... I think I've made a small lunar change before but that was a fork I believe

packages/core/src/components/Timeline/SimpleItem.tsx
Tgeo: Agree it is more standard with lunar practices, I'll make that update

packages/core/src/components/Timeline/styles.ts
Tgeo: I'll see if I can remove but I think it will be required since the "secondary" item style actually has no `children` prop - so the height of the wrapper here is actually just providing the default "empty space" below the secondary item's title

packages/core/src/components/Timeline/styles.ts
Tgeo: Nice!

packages/core/src/components/Timeline/styles.ts
Tgeo: Looks like I was able to clean it up some, good suggestion!

PR #322, opened 2020-02-19
new(Timeline): Add Timeline component

Tgeo:

## Description

This is a clone of https://github.com/airbnb/lunar/pull/321 which was closed because it was based off a fork. This PR _does_ include the addressed feedback from that PR (I've left comments on the previous PR if there is interest).

Add a new component, `<Timeline />` which shows a chronological set of events.

Open to thoughts on implementation. I decided to go with two different types of "child" items instead of creating one child item component that would have its behavior controlled via boolean props.

Top-level "Timeline" component is empty (just renders `children`) - but seems appropriate for general standardization and for future proofing.

## Motivation and Context

This component already has an implementation internally (Nova) but will be used by other apps, so it makes sense to genericize it for other consumers.

## Testing

Tested locally in storybook.

## Screenshots

image

## Checklist


packages/core/src/components/Timeline/Item.tsx
```diff
@@ -0,0 +1,66 @@
...
+          </div>
+        }
+      >
+        {secondary ? (
+          children
+        ) : (
+          <Text small muted>
+            <T
+              phrase="%{mediumDate} (%{relativeDate})"
+              k="lunar.timeline.date"
```
Tgeo: Unsure if this was how we wanted to use `<T />` ... since there isn't any actual verbiage I went with this. Also, regarding the `key` (`k` prop) -- I just made up that string? Any guidance?

Tgeo:
@stefhatcher @milesj updated and no longer on a fork - PTAL

packages/core/src/components/Timeline/styles.ts
```diff
@@ -0,0 +1,24 @@
...
+  wrapper: {
+    marginLeft: unit * 0.5 - 1,
+    borderLeft: ui.borderThick,
+    paddingLeft: unit * 2.5,
+    paddingBottom: unit * 7,
+  },
+
+  wrapper_secondary: {
+    paddingBottom: 0,
+    height: unit * 2.75,
```
stefhatcher: can this be minHeight? or not at all - use paddingBottom to make up the height difference or something? what if the translation is really long and the line wraps, or the line wraps in general? how does this height interact?

Tgeo: @stefhatcher so for secondary items, we don't render the `children` prop in the "content" area of the timeline item - it gets rendered as the timeline item "title" - for the "secondary item" scenario - the wrapper will just be an _empty_ `<div />` - and we've provided this height in order to give it a minimum spacing between items - the screenshot should show the empty space I'm talking about  
image

I set the `children` prop for a `secondary` item here to wrap, it looks OK to me - LMK if I've misunderstood

Tgeo: FWIW it _could_ be `min-height` if thats preferred

Tgeo:
Approved Happo CI

packages/core/src/components/Timeline/styles.ts
stefhatcher: @Tgeo i hope you dont mind, but i was very intrigued by the secondary wrapping problem (wanted to preserve the spacing around the dot, while allowing the secondary content to wrap with the bar growing) and re-wrote this to use css grid 

https://github.com/airbnb/lunar/pull/322/commits/089344d899c8d7c20eeeb8e4b3b9fa35eeb29950

and pushed it up. thoughts?

in this screenshot the top is the new styles with wrapping secondary, with the previous below to compare (i also fixed the off by one pix in the screenshot, so ignore that part):
Screen Shot 2020-02-20 at 7 56 29 AM

Tgeo: @stefhatcher don't mind at all - I wish we did that more often

Tgeo:
@stefhatcher I changed the column gap slightly on top of your changes. Given your changes, is there anything outstanding for this PR?

PR #326, opened 2020-02-24
breaking(DataTable): Remove edit mode, header buttons, and selectable state.

schillerk:

## Motivation and Context
Selectable is now an anti-pattern I've had to caution many people against using. Having the DataTable own it's own selectable state is a nightmare that often leads to redundant state that has to be synced manually, improperly selected rows when data is modified, formatting errors, and more.

EditMode isn't used, and adds a lot of complexity. It's has some nice logic to handle stuff like reverting edits, but really ought to be handled outside DataTable.

These are breaking changes intended to be merged into the v3 release.

## Screenshots
DataTable has always been formatted to work without selectable, so this doesn't really change. Expandable spacing is still correct:

Screen Shot 2020-02-24 at 3 25 06 PM

## Checklist


packages/composer/src/components/Composer.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useCallback, useState, useReducer } from 'react';
 import useStyles from '@airbnb/lunar/lib/hooks/useStyles';
 import FormErrorMessage from '@airbnb/lunar/lib/components/FormErrorMessage';
+import uuid from 'uuid/v4';
```
kristw: Is `/v4' required?

packages/composer/src/components/Footer/index.tsx
```diff
@@ -28,12 +28,18 @@ export default function Footer() {
       {tips.map(tip => (
         <Tip key={tip.name || tip.combo}>
           <Mark>
-            {tip.combo.split('+').map((symbol, i) => (
-              <span key={symbol}>
-                {i !== 0 && ' + '}
-                <Symbol char={SYMBOLS[symbol] || symbol.toLowerCase()} />
-              </span>
-            ))}
+            {tip.combo.split('+').map((symbol, i) => {
```
kristw: Could use `.filter(symbol => !symbol.startsWith('!'))` instead of `if () { return null }`

packages/core/src/components/DataTable/ColumnLabels.tsx
```diff
@@ -48,12 +47,11 @@ export default function ColumnLabels({
   rowHeight?: RowHeightOptions;
   columnHeaderHeight?: HeightOptions;
   expandable?: boolean;
-  selectable?: boolean;
   columnMetadata?: ColumnMetadata;
   columnLabelCase?: ColumnLabelCase;
 }) {
   return ({ className, columns, style }: ColumnLabelsProps) => {
-    const leftmostIdx = Number(expandable) + Number(selectable);
+    const leftmostIdx = Number(expandable);
```
ktmud: I would make it explicit `const leftMostIdx = expandable ? 1 : 0` if the addition magic is not needed anymore.

packages/core/src/components/DataTable/DataTable.tsx
```diff
@@ -518,11 +310,10 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
             overscanRowCount={
               dynamicRowHeight && showAllRows ? expandedData.length : overscanRowCount!
             }
-            onRowClick={this.handleRowClick}
+            onRowClick={() => rowClickCallback}
```
ktmud: Should this be  `onRowClick={() => rowClickCallback()}`? And was it intentional to hide the onRowClick API from `react-virtualized`? It's [function signature](https://github.com/bvaughn/react-virtualized/blob/master/docs/Table.md) is `({ event: Event, index: number, rowData: any }): void` and I imagine at least some of that would be useful to the end users.

milesj:
@schillerk This ready to review?

packages/core/src/components/DataTable/DataTable.tsx
milesj: ^^^

I'd also keep `handleRowClick` to keep referential equality.

PR #327, opened 2020-02-26
fix(Timeline): Correctly import IconRecord and export Timeline/Item

Tgeo:

## Description

- Import path for `IconRecord` was incorrect - works locally and in storybook but not when consuming the npm package
- Exported `Timeline/Item` from index. This seems to confuse Storybook and the props for `Timeline/Item` are shown when viewing `Timeline` itself.
- **Disable happo for `Timeline` component**, see: https://github.com/airbnb/lunar/pull/327#issuecomment-591682052

## Motivation and Context

Can't use `Timeline` without `IconRecord` import change

## Testing

Local storybook

## Screenshots

No visual change

## Checklist

### DIFF
--- a/packages/core/src/components/Timeline/Item.tsx
+++ b/packages/core/src/components/Timeline/Item.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import IconRecord from '@airbnb/lunar-icons/src/interface/IconRecord';
+import IconRecord from '@airbnb/lunar-icons/lib/interface/IconRecord';
 import useStyles from '../../hooks/useStyles';
 import { styleSheetItem } from './styles';
 import Text from '../Text';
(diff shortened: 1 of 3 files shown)


Tgeo:
Going to fix happo diff issue with a hardcoded date so this won't cause a visual diff for every PR...

Tgeo:
> Going to fix happo diff issue with a hardcoded date so this won't cause a visual diff for every PR...

Actually I'm going to have to disable happo for `Timeline` since it uses a relative date ... I can't see a way it won't _always_ render a visual diff

stefhatcher:
a fixed time instead of `now` would fix this issue and we could have happo diffs.

Tgeo:
@stefhatcher wouldn't the relative time ("A few days ago") still be problematic since even a hardcoded date will change relative to the current time

PR #329, opened 2020-03-02
style: Update shadows to new style.

milesj:

## Description

Updates our shadows to:

Large `0px 8px 28px rgba(0, 0, 0, 0.28)`
Default `0px 6px 20px rgba(0, 0, 0, 0.2)`
Small `0px 4px 16px rgba(0, 0, 0, 0.12)`

## Motivation and Context

Just part of the new design.

## Checklist

### DIFF
--- a/packages/core/src/components/Lightbox/index.tsx
+++ b/packages/core/src/components/Lightbox/index.tsx
@@ -140,7 +140,7 @@ export default class Lightbox extends React.PureComponent<LightboxProps, Lightbo
     );
 
     return (
-      <Sheet compact noAnimation portal visible header={header} onClose={onClose}>
+      <Sheet compact headerShadow noAnimation portal visible header={header} onClose={onClose}>
         <LightboxImage
           aside={aside}
           alt={alt}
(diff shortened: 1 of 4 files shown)


packages/core/src/components/Lightbox/index.tsx
```diff
@@ -140,7 +140,7 @@ export default class Lightbox extends React.PureComponent<LightboxProps, Lightbo
     );
 
     return (
-      <Sheet compact headerShadow noAnimation portal visible header={header} onClose={onClose}>
```
stefhatcher: instead of the shadow, add a border bottom to the header

milesj: Switched it to a thick border.

packages/core/src/components/Sheet/styles.ts
```diff
@@ -147,7 +147,7 @@ export const styleSheetSheet: StyleSheet = ({ color, pattern, ui, unit }) => ({
   },
 
   headerShadow: {
-    boxShadow: ui.boxShadow,
+    borderBottom: ui.borderThick,
```
stefhatcher: why thick?

milesj: It's the same height to what the shadow was. IMO, I like it.

PR #330, opened 2020-03-02
fix(Aside): Clean up side-effects.

milesj:

## Description

Clean up how side-effects work, with stories to back them up.

## Motivation and Context

This causes a circular loop as they both affect each other.

## Checklist

### DIFF
--- a/packages/layouts/src/components/Aside/index.tsx
+++ b/packages/layouts/src/components/Aside/index.tsx
@@ -92,15 +92,13 @@ export default function Aside({
 
   useEffect(() => {
     if (expanded !== undefined) {
-      setExpanded(prev => {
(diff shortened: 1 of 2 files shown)


packages/layouts/src/components/Aside/index.tsx
```diff
@@ -99,12 +99,8 @@ export default function Aside({
 
         return expanded;
       });
-
-      if (onCollapseToggle) {
-        onCollapseToggle(expanded);
```
stefhatcher: i think you can make it work where this is still supported

milesj:
@stefhatcher Updated. If using both in parallel, you'll have to manually sync things to work correctly.

PR #334, opened 2020-03-07
update(Tabs) add customPushState prop

gaarf:

## Description

add a new optional `customPushState` prop to `<Tabs />` to be used in conjunction with `persistWithHash`.

## Motivation and Context

One may want to use _react-router_'s `pushState` instead of the native one.

## Testing

Added to existing spec

## Screenshots

n/a

## Checklist

### DIFF
--- a/packages/core/src/components/Tabs/index.tsx
+++ b/packages/core/src/components/Tabs/index.tsx
@@ -19,6 +19,8 @@ export type TabsProps<T extends string = string> = {
   onChange?: (key: T) => void;
   /** Persist the selected tab through the defined URL hash. */
   persistWithHash?: string;
+  /** A function to use in lieu of the native pushState. */
+  customPushState?: typeof history.pushState;
   /** Secondary tab style, implies borderless. */
   secondary?: boolean;
   /** Wrap tabs in a scrollable region. */
@@ -45,6 +47,7 @@ function Tabs<T extends string = string>({
   scrollable,
   stretched,
   persistWithHash,
+  customPushState,
   defaultKey,
   onChange,
   styleSheet,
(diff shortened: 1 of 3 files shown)


milesj:
@gaarf We don't support forks anymore, so can you get added to the Airbnb org, then you can just create an actual branch.

packages/core/src/components/Tabs/index.tsx
```diff
@@ -19,6 +19,8 @@ export type TabsProps<T extends string = string> = {
   onChange?: (key: T) => void;
   /** Persist the selected tab through the defined URL hash. */
   persistWithHash?: string;
+  /** A function to use in lieu of the native pushState. */
+  customPushState?: typeof history.pushState;
```
milesj: Maybe `onPushState` since its a handler.

alecklandgraf:
What do you think about passing in a history object?

gaarf:
There are many different « history objects » that one could be using, ie
not everyone uses react-router. IMHO it’s better to leave this optional
implementation to the component consumer. For example this will for example
make it possible to call replaceState instead of pushState.

Le dim. 8 mars 2020 à 3:02 PM, Aleck Landgraf <notifications@github.com> a
écrit :

packages/core/src/components/Tabs/index.tsx
stefhatcher: +1 onPushState

alecklandgraf:
Lunar only uses `history` here, so I'm OK with this. Allowing a `history` object has the benefits of not needing to pass any other history method in later. 

Note react-router lets you select the history object, all should have a similar interface to `typeof history` 

https://github.com/ReactTraining/history/blob/3f69f9e07b0a739419704cffc3b3563133281548/modules/index.js#L1-L3

gaarf:
closing in favor of https://github.com/airbnb/lunar/pull/335

PR #337, opened 2020-03-11
update(Tooltip): Add `toggleOnClick`  prop.

apiel51:

## Description
Add `toggleOnClick` prop that toggles the display of the tooltip content on user click when true.

## Motivation and Context
Request from design

## Testing
Local

## Screenshots

toggleonclick

## Checklist

### DIFF
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -32,6 +32,8 @@ export type TooltipProps = {
   onShow?: () => void;
   /** True to prevent dismissmal on mouse down. */
   remainOnMouseDown?: boolean;
+  /** True to toggle tooltip display on click.  */
+  toggleOnClick?: boolean;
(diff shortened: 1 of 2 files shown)


milesj:
I believe the reason we haven't done this yet is there is no treatment to denote that there is a tooltip on click. This kind of has similar issues. Has there been any design input?

stefhatcher:
is it a tooltip if it can be toggled open/closed? or is that more of a popover?

apiel51:
@milesj In the design that uses this functionality the child of the tooltip is a button, so I think it would be fairly intuitive for users to click it. In other situations, however, it may not be so clear. Maybe when this prop is true we could override the cursor on hover so it's a pointer, which makes clicking intuitive even if the child element is not a button?

@stefhatcher I think "popover" is also an accurate term, but I figured this functionality shared so much with `Tooltip` that it made sense to use a prop instead of creating a new component.

PR #342, opened 2020-03-19
fix(GradientScroller): Initially hide gradient end.

gaarf:

## Motivation and Context

When using `<GradientScrollbar showArrows />`, if the content does _not_ overflow, we would see a flash of the arrow before the ResizeObserver does it's thing.

## Testing

StyleBook

## Screenshots

before this change, in nova usage of `<Tabs scrollable />` :

demo

## Checklist

### DIFF
--- a/packages/core/src/components/GradientScroller/story.tsx
+++ b/packages/core/src/components/GradientScroller/story.tsx
@@ -33,7 +33,7 @@ export default {
   title: 'Core/GradientScroller',
   parameters: {
     inspectComponents: [GradientScroller],
-    happo: { delay: 1000 },
+    happo: { delay: 500 },
   },
 };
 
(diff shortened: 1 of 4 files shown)


stefhatcher:
i think this breaks tabs scrollable default state (seen in happo diffs):

Screen Shot 2020-03-19 at 4 22 13 PM

can you confirm the arrow shows in the story `With scrollable variable height tabs.`? if it does, might need to add a delay for happo to render the arrow

gaarf:
@stefhatcher thanks for looking! I checked and added a delay.

Screen Shot 2020-03-19 at 7 42 16 PM

packages/core/src/components/GradientScroller/story.tsx
```diff
@@ -70,6 +71,18 @@ withLeftAndRightArrows.story = {
   name: 'With left and right arrows.',
 };
 
+export function withArrowsButNotShown() {
+  return (
+    <GradientScroller showArrows hideScrollbar>
+      <Text>{"Hello"}</Text>
```
stefhatcher: `{"Hello"}` is throwing some prettier/lint errors. update this with `<LoremIpsum short />` please (we try to use that for story copy)

gaarf:
@stefhatcher PTAL 🙏

PR #357, opened 2020-04-18
new(Tooltip): Add optional accessibilityLabel for perf

williaster:
cc @avand @stopachka

## Description

This PR adds an optional `accessibilityLabel` prop to the `Tooltip`, mostly to enable an opt-in (non-breaking) perf improvement. 

## Motivation and Context

Currently the `content` prop of `Tooltip` is rendered in an invisible `Portal` for a11y, to serve as an `aria-labeledby` element that is rendered whether or not the `Tooltip` popup is visible. This can lead to performance issues on pages with 100s of `Tooltips` as `content` is rendered at all times.

We discussed that one solution to ensure accessibility but allow for optimization is to override this behavior by specifying an `aria-label` (`accessibilityLabel` prop) to be used instead. 

Note: there was some discussion about whether omission of `aria-labelledby`/duplication of content should be conditional on whether `props.toggleOnClick === false` as you could potentially have functionality in `content` that would not be captured with a simple label. After thinking about this more I think it could be confusing to the developer but am still open to adding it if you want.

## Testing

- [x] CI
- [x] functional – `aria-label` and `aria-labelledby` work in new story

## Screenshots

New example
image

## Checklist

### DIFF
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -18,7 +18,7 @@ const EMPTY_TARGET_RECT: ClientRect = {
 };
 
 export type TooltipProps = {
-  /** Accessibility label. If not specified, all tooltip content is duplicated in an accessibility portal rendered at all times. */
+  /** Accessibility label. If not specified, all tooltip content is duplicated, rendered in an off-screen element with a separate layer. */
   accessibilityLabel?: string;
   /** Inline content to hover. */
   children: NonNullable<React.ReactNode>;
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -18,6 +18,8 @@ const EMPTY_TARGET_RECT: ClientRect = {
 };
 
 export type TooltipProps = {
+  /** Accessibility label. If not specified, all tooltip content is duplicated in an accessibility portal rendered at all times. */
+  accessibilityLabel?: string;
   /** Inline content to hover. */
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Tooltip/story.tsx
```diff
@@ -214,3 +214,33 @@ export function toggleWithClick() {
...
   name: 'Toggle tooltip on click',
 };
+
+export function withAccessibilityLabel() {
+  return (
+    <>
+      <Spacing bottom={1}>
+        <Tooltip
+          accessibilityLabel="Override accessibility content"
+          content="I have an aria-labelledby instead of aria-label"
```
stopachka, line 224: noob q: is this the other way arround? 

(i.e here because we added accessibelityLabel, this _has_ an aria-label instead of an aria-labelledby)

williaster: yes! good catch 😅 I swapped these around a couple times and tho I got the `children` text right I failed to swap the `content` 🙏

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -18,6 +18,8 @@ const EMPTY_TARGET_RECT: ClientRect = {
 };
 
 export type TooltipProps = {
+  /** Accessibility label. If not specified, all tooltip content is duplicated in an accessibility portal rendered at all times. */
```
stefhatcher, line 21: small nit: to change the last part of the sentence to `is duplicated, rendered in an off-screen element with a separate layer.`

PR #361, opened 2020-04-22
update(ErrorMessage): Add trace ID link

Tgeo:

## Description

- Add `traceURL` setting
- Use `trace_id` property from `ErrorMessage` for link for more info on that trace ID
- Update the `Button` to a `Link` (no need for it to be a button that I can think of - don't need to generate the URL "on the fly" when the button is clicked, we can use a plain anchor tag)

## Motivation and Context

Allow better debugging when a `trace_id` is present in an `Error` passed to `ErrorMessage` component.

## Testing

Added tests. Ran `sg` locally and verified.

## Screenshots

image

## Checklist

### DIFF
--- a/packages/core/src/components/ErrorMessage/index.tsx
+++ b/packages/core/src/components/ErrorMessage/index.tsx
@@ -25,11 +25,6 @@ export function getErrorMessage(error: string | ErrorType, includeCode: boolean 
   return message || '';
 }
 
-// istanbul ignore next
-function createRedirectURL(id: string, url?: string) {
-  return () => window.open(url || Core.settings.errorURL.replace('{{id}}', id), '_blank');
-}
-
 export type ErrorMessageProps = {
   /** An `Error` instance or an API endpoint response. */
   error?: ErrorType;
@@ -57,8 +52,8 @@ export default function ErrorMessage({
 
   const message = subtitle || getErrorMessage(error);
   const code = error.error_code || '';
-  const id = error.error_id || '';
-  const url = error.error_url || '';
+  const errorID = error.error_id || '';
+  const traceID = error.trace_id || '';
 
   if (inline) {
     return <StatusText danger>{message}</StatusText>;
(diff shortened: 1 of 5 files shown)


packages/core/src/components/ErrorMessage/index.tsx
```diff
@@ -72,13 +67,29 @@ export default function ErrorMessage({
...
       )}
+
+      {traceID && (
+        <Spacing top={1}>
+          <MutedButton
+            inverted
+            openInNewWindow
+            href={Core.settings.traceURL.replace('{{id}}', traceID)}
+          >
+            <T k="lunar.trace.viewDetails" phrase="View trace details" />
```
Tgeo, line 89: Do I need to do anything with this new phrase?

packages/core/src/components/ErrorMessage/index.tsx
```diff
@@ -25,11 +25,6 @@ export function getErrorMessage(error: string | ErrorType, includeCode: boolean
   return message || '';
 }
 
-// istanbul ignore next
-function createRedirectURL(id: string, url?: string) {
```
Tgeo, line 29: Seemed unnecessary to generate the URL on-the-fly on a Button's `onClick` when we can just set an href

packages/core/src/components/ErrorMessage/index.tsx
milesj: Yes you need to add it internally like our other apps.

packages/core/src/components/ErrorMessage/index.tsx
```diff
@@ -72,13 +67,29 @@ export default function ErrorMessage({
...
+            openInNewWindow
+            href={error.error_url || Core.settings.errorURL.replace('{{id}}', errorID)}
+          >
             <T k="lunar.error.viewDetails" phrase="View error details" />
           </MutedButton>
         </Spacing>
       )}
+
+      {traceID && (
+        <Spacing top={1}>
```
milesj, line 83: Buttons shouldn't be stacked. Wrap with `ButtonGroup` instead.

packages/core/src/components/ErrorMessage/index.tsx
```diff
@@ -72,13 +67,29 @@ export default function ErrorMessage({
...
+          <MutedButton
+            inverted
```
milesj, line 85: Can you also make them `small`.

packages/core/src/components/ErrorMessage/index.tsx
Tgeo: Done

Tgeo:
@milesj feedback addressed

PR #362, opened 2020-04-23
new(Range): Add input range component

williaster:
cc @kenchendesign 

## Description

This PR adds a new form element `Range` to both the `core` and `forms` packages. Functionality:
- can specify `min`/`max`/`step` (typical range values)
- can specify `width` (in `px`)
- can annotate specific values you choose, with any (optional) labels you want (see stories below)
- can show value in a tooltip on hover, or persistently; can also customize the rendering of the value

## Motivation and Context

See solar #938 for mocks (note the double-slider variant has not been implemented in this initial version, this would need to function similarly to `MultiComplete` and adds complexity). 

Notes on implementation (major kudos to @stefhatcher for very helpful discussions! 🙌):
- After extensive discussion about the technical implementation (`div`s vs `input`s) I found some work arounds for implementing with a proper `input` which is better for accessibility and simplified integration with the `forms` package + parity with the other form components. 
- The only downside of the `input` approach was that I had to do everything in `px` space (not `%`) to achieve the desired style customizations. The main challenge was that the **edges** (not the _center_) of the slider handle align with the **edges** of the slider track which means the total width of the slider is not `100%` but `100% - HANDLE_SIZE_PX`. Without an offset correction this meant that labels and annotations are placed incorrectly across the entire track:

- My solution to the above was to use a `background` `gradient` (which we already needed for filling the slider up to the selected value) to inset the edges (note the offset is still needed):

- Styles are only specified for webkit, it doesn't look quite as nice in firefox or safari but I think is totally acceptable (see below)

## Testing
- [x] CI
  - [x] new tests in `core` package
  - [x] new tests in `forms` package 
- [x] `core` package stories are functional
- [x] `forms` package stories are functional / the value is returned properly on form submission
- [x] disabled state works as expected
- [x] can be tab-ed to and updated with keyboard
- [x] tooltips show with keyboard control
- [x] offsets are correct for negative min/max values
- [x] storybook accessibility checks pass – still verifying

## Screenshots

**A standard range slider**

**With annotations and step size**

**With tooltips**

**With custom width and min/max/step**

**In Firefox**

**In Safari**

**Forms package**

## Checklist

### DIFF
--- a/packages/core/src/components/Range/index.tsx
+++ b/packages/core/src/components/Range/index.tsx
@@ -1,0 +1,22 @@
+import React, { useState } from 'react';
+import { v4 as uuid } from 'uuid';
+import FormField, { FormFieldProps, partitionFieldProps } from '../FormField';
+import BaseInputRange, { BaseInputRangeProps } from '../private/BaseInputRange';
+import { IgnoreAttributes } from '../private/FormInput';
+
+export type RangeProps = Omit<BaseInputRangeProps, 'id'> &
+  FormFieldProps &
+  Omit<React.InputHTMLAttributes<HTMLInputElement>, IgnoreAttributes>;
+
+/** A controlled range input. */
+export default function Range(props: RangeProps) {
+  const { fieldProps, inputProps } = partitionFieldProps(props);
+  const [id] = useState(() => uuid());
+
+  return (
+    <FormField {...fieldProps} id={id}>
+      {/** inputProps.value is a string, we need a number */}
+      <BaseInputRange {...inputProps} value={props.value} id={id} />
+    </FormField>
+  );
+}
--- a/packages/core/src/components/Range/index.tsx
+++ b/packages/core/src/components/Range/index.tsx
@@ -1,10 +1,10 @@
 import React, { useState } from 'react';
 import { v4 as uuid } from 'uuid';
 import FormField, { FormFieldProps, partitionFieldProps } from '../FormField';
-import BaseInputRange, { BaseInputRangeProps } from '../private/BaseInputRange';
+import BaseRange, { BaseRangeProps } from './BaseRange';
 import { IgnoreAttributes } from '../private/FormInput';
 
-export type RangeProps = Omit<BaseInputRangeProps, 'id'> &
+export type RangeProps = Omit<BaseRangeProps, 'id'> &
   FormFieldProps &
   Omit<React.InputHTMLAttributes<HTMLInputElement>, IgnoreAttributes>;
 
@@ -15,8 +15,8 @@ export default function Range(props: RangeProps) {
 
   return (
     <FormField {...fieldProps} id={id}>
-      {/** inputProps.value is a string, we need a number */}
-      <BaseInputRange {...inputProps} value={props.value} id={id} />
+      {/** inputProps.value is typed as a string, BaseRange takes a number. */}
+      <BaseRange {...inputProps} value={props.value} id={id} />
     </FormField>
   );
 }
--- a/packages/core/src/components/Range/styles.ts
+++ b/packages/core/src/components/Range/styles.ts
@@ -36,7 +36,7 @@ export const stylesheetInputRange: StyleSheet = ({ unit, ui, color, font, patter
           transition: `opacity ${ui.transitionTime}, backgroundColor ${ui.transitionTime}`,
           width: HANDLE_SIZE,
           height: HANDLE_SIZE,
-          border: `2px solid ${color.accent.borderActive}`,
+          border: `2px solid ${color.core.primary[3]}`,
           borderRadius: HANDLE_SIZE,
           backgroundColor: color.accent.bg,
         },
@@ -44,10 +44,12 @@ export const stylesheetInputRange: StyleSheet = ({ unit, ui, color, font, patter
         ':not([disabled]):active::-webkit-slider-thumb, :not([disabled]):focus::-webkit-slider-thumb': {
           ...pattern.themedFocus,
           backgroundColor: color.accent.bgHover,
+          borderColor: color.core.primary[4],
         },
 
         ':not([disabled]):hover::-webkit-slider-thumb': {
           backgroundColor: color.accent.bgHover,
+          borderColor: color.core.primary[4],
         },
       },
     },
--- a/packages/core/src/components/Range/styles.ts
+++ b/packages/core/src/components/Range/styles.ts
@@ -1,4 +1,4 @@
-import { StyleSheet } from '../../../hooks/useStyles';
+import { StyleSheet } from '../../hooks/useStyles';
 
 export const HANDLE_SIZE = 18;
 export const HALF_HANDLE_SIZE = HANDLE_SIZE / 2;
--- a/packages/core/src/components/private/BaseInputRange/index.tsx
+++ b/packages/core/src/components/private/BaseInputRange/index.tsx
@@ -40,7 +40,7 @@ export type BaseInputRangeProps = {
   /** Step size for the range. */
   step?: number;
   /** Current value. */
-  value: number;
+  value?: number;
   /** Width of the input. */
   width?: number;
 };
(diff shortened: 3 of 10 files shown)


packages/core/src/components/Range/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+/** A controlled range input. */
+export default function Range(props: RangeProps) {
+  const { fieldProps, inputProps } = partitionFieldProps(props);
+  const [id] = useState(() => uuid());
+
+  return (
+    <FormField {...fieldProps} id={id}>
+      {/** inputProps.value is a string, we need a number */}
+      <BaseInputRange {...inputProps} value={props.value} id={id} />
```
williaster, line 19: I wasn't sure of a cleaner way around the `value` type problem. It's much easier to use `number` in the component since we have to do so many calculations

packages/core/src/components/private/BaseInputRange/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+    position: 'relative',
+    height: 3 * unit,
+  },
+
+  container_disabled: {
+    opacity: 0.4,
+  },
+
+  input: {
+    transform: 'translateY(-2px)',
```
williaster, line 20: I really couldn't figure out what determined the slider height in the browser, so this is empirical 😭

packages/core/src/components/private/BaseInputRange/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+  input_disabled: {
+    '@selectors': {
+      ':hover::-webkit-slider-thumb': { boxShadow: `0 0 0 2px ${color.accent.borderActive}` },
+    },
+  },
+
+  annotation: {
+    position: 'absolute',
+    top: 1.25 * unit + 1, // align in track center
+    width: TRACK_HEIGHT,
```
williaster, line 61: will update these to `MARK_SIZE`

packages/core/src/components/private/BaseInputRange/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  input_disabled: {
+    '@selectors': {
+      ':hover::-webkit-slider-thumb': { boxShadow: `0 0 0 2px ${color.accent.borderActive}` },
+    },
+  },
+
+  annotation: {
+    position: 'absolute',
+    top: 1.25 * unit + 1, // align in track center
```
williaster, line 60: will update to `TOTAL_HEIGHT - 1` so it's in sync with the height of the container

packages/core/src/components/private/BaseInputRange/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+    if (this.props.showTooltip && !this.props.alwaysShowTooltip) {
+      this.setState({ showPopup: false });
+    }
+  };
+
+  renderTooltip(leftOffset: number) {
+    const { renderTooltipContent, value = 0, cx, styles } = this.props;
+    return (
+      <div
+        role="tooltip"
```
williaster, line 93: I didn't think this needed the complexity / perf implications of the `Tooltip` component so implemented a simplified version (open to updating based on what you think). 

The designs have the tooltip on **top**, but I implemented on the **bottom** because if `renderTooltipContent` results in `>1` line of text, the fixed positioning on top would be wrong without more complex logic / dynamic positioning (used in `Tooltip`) 🤷‍♀️

stefhatcher:
for focused state, is it possible to add `pattern.themedFocus`? would look similar to when a radio button is focused

packages/core/src/components/private/BaseInputRange/index.tsx
stefhatcher: i think this tooltip should match the other tooltip's default state (white background, normal text) and have the option to `invert`

packages/core/src/components/private/BaseInputRange/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+import { stylesheetInputRange, HANDLE_SIZE, HALF_HANDLE_SIZE, MARK_SIZE } from './styles';
+
+/**
+ * By default, the **edges** (not the _center_) of the slider handle align with
+ * the **edges** of the slider track. Thus the total width of the slider is not
+ * `100%` but `100% - HANDLE_SIZE_PX`. The default position is correct at 50%,
+ * so we offset based on a fractional distance from the center ()
+ * position based on the percent
+ */
+const getCorrectedPosition = memoize(
```
stefhatcher, line 15: could css's `calc()` have helped with calculating `100% - HANDLE_SIZE_PX`? so it could still be % based?

packages/core/src/components/private/BaseInputRange/styles.ts
```diff
@@ -0,0 +1,108 @@
...
+    height: 3 * unit,
+  },
+
+  container_disabled: {
+    opacity: 0.4,
+  },
+
+  input: {
+    transform: 'translateY(-2px)',
+    '-webkit-appearance': 'none',
```
stefhatcher, line 21: i think `appearance` is valid too? might help to add that as well?

packages/core/src/components/private/BaseInputRange/index.tsx
williaster: I can play around with that. I think we would still potentially need the `px` calculations for knowing when to hide the annotations when they overlap with the handle (I couldn't figure out zIndex of the handle, maybe can play with that more too, now that everything else is working)

elibrumbaugh:
That custom render is amazing!

packages/core/src/components/private/BaseInputRange/index.tsx
```diff
@@ -0,0 +1,205 @@
+import React from 'react';
```
milesj, line 1: This can probably just be in `Range.tsx` since it's not used anywhere else.

packages/core/src/components/private/BaseInputRange/index.tsx
williaster: @stefhatcher I'm curious, this `calc` still wouldn't truly be `%` based since it would be a `px` offset. Are you suggesting this because it should be more performant than the current implementation?

stefhatcher: true - just a suggested if it helped clear up any of the calculation logic! could be more performant putting it on the browser, but not sure

williaster:
some pics of the style updates

**bottom is hovered**


**bottom is focused/active**


**tooltip background**
white by default, bottom is inverted

packages/core/src/components/private/BaseInputRange/index.tsx
williaster: moved to `Range/` as `BaseRange` so we can keep it separate from the `Form`-wrapped one (`Range/index`)

packages/core/src/components/private/BaseInputRange/styles.ts
williaster: removed `-webkit`-prefixed styles entirely, looked the same in chrome + firefox 🤷‍♀️

williaster:
after rebasing on latest the happo diffs lgtm

williaster:
@stefhatcher @milesj comments addressed, lmk what you think when you have time 🙏

packages/core/src/components/Range/styles.ts
```diff
@@ -0,0 +1,110 @@
...
+          width: HANDLE_SIZE,
+          height: HANDLE_SIZE,
+          border: `2px solid ${color.accent.borderActive}`,
+          borderRadius: HANDLE_SIZE,
+          backgroundColor: color.accent.bg,
+        },
+
+        ':not([disabled]):active::-webkit-slider-thumb, :not([disabled]):focus::-webkit-slider-thumb': {
+          ...pattern.themedFocus,
+          backgroundColor: color.accent.bgHover,
```
stefhatcher, line 46: sorry, thought we discussed this, but make the border darker too (like inputs and radio button borders get darker on hover/focus). same for hover styles below

williaster:
darker background + border on hover (top) vs not (bottom)

packages/core/src/components/Range/styles.ts
williaster: sorry for the oversight there, done

williaster:
and tests added to both packages 👍

williaster:
@stefhatcher I added tests for both packages and updated the border color to match button.

PR #367, opened 2020-04-30
new(DataTable): Add xLarge row size. Fix forceUpdate logic.

williaster:
cc @schillerk 

## Description

This PR
- adds a new row size to `DataTable`: `xLarge`
- improves the `rowHeight` type from `string` to a union of possible `string`
- tries to fix some conditions which lead to infinite re-renders from `forceUpdate` calls in `componentDidUpdate`

## Motivation and Context

I've been trying to upgrade to lunar v3 and some of our tables are pretty broken, especially for dynamic row height. Adding an additional fixed row height will alleviate some of these issues for us, and I'm hoping the `forceUpdate` logic may help. I'm not sure if there have been regressions, but it's pretty easy to generate a stack trace in the storybook now. Go [here](https://airbnb.io/lunar/?path=/story/core-datatable--a-table-with-dynamic-sort-key), expand the row, then sort a column 😱 

I couldn't think of a great way to remove the force updates, but I tried to clean it up a bit. You can still get the stack trace from resizing the window many times, unfortunately the `AutoSizer` doesn't support a `debounce` property 😢 

## Testing

- [x] CI
- [x] functional 
  - [x] `xLarge` row height renders as expected
  - [x] expanding a row + sorting no longer gives a stack trace

## Screenshots

**xLarge** (no story changes)
image

## Checklist

### DIFF
--- a/packages/core/src/components/DataTable/DataTable.tsx
+++ b/packages/core/src/components/DataTable/DataTable.tsx
@@ -66,6 +66,8 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
     sortDirection: this.props.sortDirectionOverride!,
   };
 
+  timeoutId: number = 0;
+
   constructor(props: DataTableProps & WithStylesProps) {
     super(props);
     if (this.props.dataTableRef) {
@@ -114,10 +116,7 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
     const { dynamicRowHeight, data, filterData, width, height, sortByCacheKey } = this.props;
     const { sortBy, sortDirection } = this.state;
     const dimensionsChanged = width !== prevProps.width || height !== prevProps.height;
-    const sortChanged =
-      sortBy !== prevState.sortBy ||
-      sortDirection !== prevState.sortDirection ||
-      sortByCacheKey !== prevProps.sortByCacheKey;
+    const sortChanged = sortByCacheKey !== prevProps.sortByCacheKey;
     const sortedData: IndexedParentRow[] = this.getData(
       data!,
       sortBy,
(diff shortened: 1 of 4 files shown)


packages/core/src/components/DataTable/DataTable.tsx
```diff
@@ -114,10 +116,7 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
     const { dynamicRowHeight, data, filterData, width, height, sortByCacheKey } = this.props;
     const { sortBy, sortDirection } = this.state;
     const dimensionsChanged = width !== prevProps.width || height !== prevProps.height;
-    const sortChanged =
-      sortBy !== prevState.sortBy ||
```
williaster, line 118: we already invoke `this.cache.clearAll();` when updating `sortBy` or `sortByDirection`, so this was doing a double update

packages/core/src/components/DataTable/constants.tsx
```diff
@@ -13,14 +13,15 @@ export const STATUS_OPTIONS = {
 };
 
 type HeightMap = {
-  [key: string]: number;
+  [key in RowHeightOptions]: number;
```
stefhatcher, line 16: nice

packages/core/src/components/DataTable/DataTable.tsx
```diff
@@ -224,10 +222,11 @@ export class DataTable extends React.Component<DataTableProps & WithStylesProps,
 
     if (dynamicRowHeight) {
       // We need to make sure the cache is cleared before React tries to re-render.
-      window.setTimeout(() => {
+      if (this.timeoutId) window.clearTimeout(this.timeoutId);
```
milesj, line 225: If a timer is already set, shouldn't that one just run instead of clearing and creating a new one?

packages/core/src/components/DataTable/types.ts
```diff
@@ -5,7 +5,7 @@ import { DataTable } from './DataTable';
 
 export type DataTableRef = (instance: DataTable) => void;
 export type TableRef = React.RefObject<Table>;
-export type RowHeightOptions = string;
+export type RowHeightOptions = 'micro' | 'small' | 'regular' | 'large' | 'xLarge' | 'jumbo';
```
milesj, line 8: Lets just call it `xlarge` for consistency.

PR #366, opened 2020-04-30
new(DateTime, DateTimeRange, DateTimeSelect, Price, PriceComparison, PriceGroup): Add empty fallback if an invalid value is provided.

stefhatcher:
to: @milesj @stefhatcher @hayes @alecklandgraf

## Description

handle bad values like `"[Hidden]"` being added to things expected timestamps or numbers representing an amount.

## Motivation and Context

caused components to crash

## Testing

- storybook
- new tests
- old tests passing

## Screenshots

Core : DateTime - Fallback when an invalid date value is provided   ⋅ Storybook 2020-04-30 12-39-34
Core : DateTimeRange - Fallback when an invalid date values are provided   ⋅ Storybook 2020-04-30 12-40-05
Core : DateTimeSelect - Fallback to today ' s date if value is invalid ⋅ Storybook 2020-04-30 12-40-29
Core : Price - Fallback when an invalid amount is provided   ⋅ Storybook 2020-04-30 12-41-09
Core : PriceComparison - Fallback when invalid amounts are provided   ⋅ Storybook 2020-04-30 12-41-27
Core : PriceGroup - Fallback when invalid amounts are provided   ⋅ Storybook 2020-04-30 12-42-00

## Checklist

- My code follows the style guide of this project.
- I have updated or added documentation accordingly.
- I have read the CONTRIBUTING document.

PR #369, opened 2020-05-05
update(Price, PriceComparison, PriceGroup): Handle invalid currency values.

stefhatcher:

## Description

if and invalid currency like`"[Hidden]"` is given, render `Empty`.

## Motivation and Context

more bad values. follow up of https://github.com/airbnb/lunar/pull/366

## Testing

- storybook
- added tests

## Checklist

### DIFF
--- a/packages/core/src/components/Price/index.tsx
+++ b/packages/core/src/components/Price/index.tsx
@@ -38,7 +38,7 @@ function Price({
   precision = 0,
   round,
   trimTrailingZeros,
-}: PriceProps) {
+}: PriceProps): JSX.Element | null {
   let amount = 0;
   let currency = baseCurrency;
   let micros = baseMicros;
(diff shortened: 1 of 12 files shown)


packages/core/src/components/Price/index.tsx
```diff
@@ -38,7 +38,7 @@ function Price({
   precision = 0,
   round,
   trimTrailingZeros,
-}: PriceProps) {
+}: PriceProps): JSX.Element | null {
```
milesj, line 41: This doesnt seem like it should be necessary?

stefhatcher: ah, might've been while i was trying something. will double check

packages/core/src/utils/formatPrice.ts
```diff
@@ -31,30 +31,38 @@ export default function formatPrice(
...
+      amountStr = `(${amountStr})`;
+    }
+
+    return amountStr;
+  } catch (error) {
+    // if (__DEV__) {
+    //   console.error(error);
+    // }
+
+    return '';
```
milesj, line 66: How much would break if we updated this to be null?

stefhatcher: @milesj a good chunk. ran into issues with `null` when doing the datetime stuff. went with `undefined` and then had to do a lot of type updates in our apps to handle that :/

milesj: I'm just worried that downstream consumers wouldn't be aware that invalid = empty string.

stefhatcher: good point! yeah, let's change it to `null` then

PR #371, opened 2020-05-06
new(Forms): Add support for arrays of fields.

RobinClowers:

## Description
In order to support dynamic forms in Tower, this PR adds the final form arrays plugin.
https://final-form.org/arrays

## Motivation and Context
My current work requires a form that allows users to dynamically add and remove fields from a form. I've looked at different ways to solve this without modifications to lunar, but wasn't able to find anything that solved all my issues.

## Testing
Looking for feedback here, I'm not sure what sort of testing is appropriate, since this is mostly a configuration change.

## Screenshots
This PR doesn't affect any existing lunar components.

## Checklist

### DIFF
--- a/packages/forms/package.json
+++ b/packages/forms/package.json
@@ -30,6 +30,7 @@
     "@types/lodash": "*",
     "airbnb-prop-types": "^2.15.0",
     "final-form": "^4.19.1",
+    "final-form-arrays": "^3.0.2",
     "lodash": "^4.17.15",
     "shallowequal": "^1.1.0",
     "uuid": "^7.0.3"
--- a/packages/forms/src/components/Form/index.tsx
+++ b/packages/forms/src/components/Form/index.tsx
@@ -16,6 +16,7 @@ import {
   FieldValidator,
   MutableState,
 } from 'final-form';
+import arrayMutators, { Mutators as ArrayMutators } from 'final-form-arrays';
 import T from '@airbnb/lunar/lib/components/Translate';
 import { getErrorMessage } from '@airbnb/lunar/lib/components/ErrorMessage';
 import FormErrorMessage from '@airbnb/lunar/lib/components/FormErrorMessage';
@@ -101,6 +102,7 @@ export default class Form<Data extends object = {}> extends React.Component<
       initialValues,
       mutators: {
         setFieldConfig: this.setFieldConfig,
+        ...arrayMutators,
       },
       onSubmit: this.handleSubmit,
       validate: this.handleValidate,
@@ -426,6 +428,8 @@ export default class Form<Data extends object = {}> extends React.Component<
           getFields: this.getFields,
           getState: this.getState,
           register: this.registerField,
+          // Since we have a known list of mutators, we can type it more specifically
+          mutators: (this.form.mutators as unknown) as ArrayMutators,
           submit: this.submitForm,
         }}
       >
--- a/packages/forms/src/types.ts
+++ b/packages/forms/src/types.ts
@@ -9,6 +9,18 @@ import {
   FieldValidator,
   Unsubscribe,
 } from 'final-form';
+import { Mutators } from 'final-form-arrays';
+
+export {
+  IsEqual,
+  FormState,
+  FieldState,
+  FieldSubscriber,
+  FieldSubscription,
+  FieldValidator,
+  Unsubscribe,
+  Mutators,
+};
 
 // value:
 // string - Autocomplete, DatePickerInput, DateTimeSelect, FileInput, Input, RadioButton, RadioButtonController, Select, TextArea, ToggleButtonController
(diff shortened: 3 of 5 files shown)


milesj:
Dynamic fields is already possible through field names like `foo[0]`. Does this not work?

Also, mutators are abstracted away and not available to consumers. We don't want to provide internals.

RobinClowers:
> Dynamic fields is already possible through field names like foo[0]. Does this not work?

I was able to add fields that way, but I don't see a way to remove them. Also, there is additional complexity here, changing the values of certain fields causes other fields to change complete, for example from a select to an autocomplete. In those cases, I need a way to remove the field and add a new one, otherwise the final form field state is wrong.

I'm curious about why we would want to hide mutators? I think they are really useful for advanced use cases like this. I suppose we could provide our own functions that wrap these mutators, but that just seems like extra layers of complexity that don't really add anything useful.

milesj:
The `unregisterOnUnmount` prop will unregister the field when the component is unmounted. That should clear all state related to that specific `name`.

RobinClowers:
There is still a sort of chicken and egg problem, without access to the state, I can't unmount the component, unless I duplicate the state outside of form, which is error prone.

Maybe I'm missing something, but this plugin is what the final form author recommends for this type of problem, and it's a very small library. What's the downside here?

stefhatcher:
you can keep track of the state within your component by using the `onStateUpdate` prop on `Form`. each input also has `onBatchChange` that's helpful for changing state based on that input changing.

RobinClowers:
> you can keep track of the state within your component

I really don't think duplicating this state is the right approach, in my experience it results in more code and more complexity. The final form array plugin is a really simple solution to this problem.

milesj:
I'd really need an example demo of what you're trying to do. Or what the data structure should look like.

RobinClowers:
Take a look at the Zendesk edit trigger form: https://airbnb.zendesk.com/agent/admin/triggers/360146796153, it's a very similar concept.

stefhatcher:
> I really don't think duplicating this state is the right approach, in my experience it results in more code and more complexity. The final form array plugin is a really simple solution to this problem.

it's how it's been done in the past for more complex, dynamic forms. you asked if you were maybe missing something, and that's it. that's generally the approach. it's keeping that dynamic, extra business logic local to the component using the form

RobinClowers:
> it's how it's been done in the past for more complex, dynamic forms

Right, but it requires significantly more code to solve the same problem. Code that keeps copies of state in sync is error prone and often a source of bugs. 

Again I ask, what's the downside of adding this library? Final form is great, and the guy who designed it also wrote this plugin, specifically to solve this problem.

milesj:
@RobinClowers Please don't link internal URLs. 

That being said, this is a very basic example.

```ts
import React, { useState } from 'react';
import Button from '@airbnb/lunar/lib/components/Button';
import Row from '@airbnb/lunar/lib/components/Row';
import Form, { Input } from '@airbnb/lunar-forms';

function Example() {
  const [fields, setFields] = useState([]);

  const handleAddField = (name) => {
    setFields(prev => [...prev, { name, value: '' }]);
  }

  const handleRemoveField = (field) => {
    setFields(prev => prev.filter(f => f !== field));
  }

  const handleSubmit = data => {
    // Do something with form state
  };

  return (
    <Form onSubmit={handleSubmit}>
      {fields.map(field => (
        <Row after={<Button onClick={() => handleRemoveField(field)}>Remove</Button>}>
          <Input unregisterOnUnmount name={`field[${field.name}]`} defaultValue={field.value} />
        </Row>
      ))}

      <Button onClick={() => handleAddField('test')}>Add field</Button>
    </Form>
  )
}
```

The list of fields to render should be managed outside of the form. This is literally just a list of names, what type of input, and an optional value. You _do not_ need all the form/field state that final form provides.

When the fields list is mutated, all the components will be rendered, either registering (mount) or unregistering (unmount) fields. When the form is submitted, you would have raw form/field state to interact with.

Also, the field names can be whatever you want to model this. `foo[nested][field]` (nested object), `list[0]` (list of items), `list[1]`, etc. Final form knows how to manage it all.

RobinClowers:
> This is literally just a list of names, what type of input, and an optional value.

That wouldn't work for my form, I would at least need to store an object with the name of each of the three fields in each row. Why put that burden on each form, when we have a good solution already?

milesj:
Sounds like each "field" is just a "row with fields", which also isn't hard to model.

I'm also curious how you would render the fields on your end, because this change doesn't really reflect that well (nor does FF docs).

RobinClowers:
The corresponding react package has a `<FieldArray>` component: https://github.com/final-form/react-final-form-arrays. I was just planning to call `getFields` on the form context, which does have the downside of re-rendering the whole form when any field changes. This form isn't particularly large, so I'm not worried about that.

milesj:
Lock changes are missing. Ill approve it once you `yarn install`.

PR #375, opened 2020-05-27
fix(Forms) Allow defaultValue to override initialValues

RobinClowers:

## Description
Right now, mounting a new field with a default value does nothing if an `initialValues` entry was provided for the field. This PR changes that logic so that field default value is used if present, otherwise it falls back to the initial value.

## Motivation and Context
Imagine an edit form, where you need to set `initialValues` to populate the existing data. Then the user changes the value of a field, which changes the options in a related select. To set the new default value, for that select, we need to be able to override the initial value.

## Testing
I've written a unit test that verifies the new behavior. My bigger concern is whether this is going to break an existing clients. I've been auditing usages of `initialValues` and `defaultValue` in source graph, and I'll report on what I find here.

## Checklist

### DIFF
--- a/packages/forms/src/components/Form/index.tsx
+++ b/packages/forms/src/components/Form/index.tsx
@@ -342,7 +342,8 @@ export default class Form<Data extends object = {}> extends React.Component<
   ) {
     const field = fields[name];
     const initial = getIn(formState.initialValues!, name);
-    const value = typeof initial === 'undefined' ? config.defaultValue : initial;
+    const newInitialValue = initial ?? config.defaultValue;
+    const value = config.defaultValue ?? initial;
 
     if (!field) {
       return;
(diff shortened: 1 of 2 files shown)


milesj:
Initial values was going to be deprecated, but this change seems fine. I don't have access anymore so you'll need to ping someone else.

alecklandgraf:
@RobinClowers can you check the multi complete component in story book for your branch. Happo is reporting a discrepancy. 

image

williaster:
It looks like this breaks the default values for `multicomplete`

williaster:
@RobinClowers for your use case can you not just set a `key` prop to force-update the form when  you need the initial value to change?

I don't think we should merge this until it's been tested more.

williaster:
from an API point of view I would also expect any `initialValues` that I pass to override whatever the `defaultValue` is. `initialValues` seems more specific than `defaultValue`, not vice versa.

RobinClowers:
Agreed, I'll hold off on merging. The `key` idea is interesting, but I don't think it will help; the component is re-rendering, but the initialValue will always take precedence over the defaultValue.

williaster:
> the component is re-rendering, but the initialValue will always take precedence over the defaultValue.
sorry my suggestion wasn't clear, could you update the `initialValue` coincident with the `key` change?

RobinClowers:
> initialValues seems more specific than defaultValue, not vice versa.

I'm not sure I agree, there  should be a way to override the initial value specified on the form at the field level. Otherwise, I would have to somehow update the initial value at the form level, which would be really cumbersome.

RobinClowers:
> could you update the initialValue coincident with the key change?

Ah that makes more sense. I don't know how we would get the state we need at the form level without duplicating the internal form state in the component that renders the form.

This kicks off from a field's onChange, and we need to re-render a different field with a new default value. So to do what you suggested, we would have to expose a function to the fields that allows the initialValues to be updated. But since they come from an api response initially, we would have to stick it in state and keep that in sync.

RobinClowers:
After further discussion with @williaster, we came up with a better approach. I'm closing this, I'll open a new PR shortly.

PR #376, opened 2020-05-27
update(Forms): Use form values if present on register

RobinClowers:

## Description
This PR checks for an existing field value when a field is registered, and uses that value rather than the form initial value.

## Motivation and Context
Using final form mutators, it's possible to set a field value before the initial mount of the component. I'm working on a form that uses [Final Form Arrays](
https://final-form.org/docs/final-form-arrays/getting-started) to arbitrary numbers of rows to a form. 

## Testing
I've added a unit test that verifies the new behavior

## Checklist

### DIFF
--- a/packages/forms/src/components/Form/index.tsx
+++ b/packages/forms/src/components/Form/index.tsx
@@ -342,7 +342,9 @@ export default class Form<Data extends object = {}> extends React.Component<
   ) {
     const field = fields[name];
     const initial = getIn(formState.initialValues!, name);
-    const value = typeof initial === 'undefined' ? config.defaultValue : initial;
+    const value =
+      getIn(formState.values!, name) ??
+      (typeof initial === 'undefined' ? config.defaultValue : initial);
 
     if (!field) {
       return;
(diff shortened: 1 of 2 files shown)


RobinClowers:
I don't have access to Happo, can someone please try re-running it?

williaster:
@RobinClowers I don't see any way to re-run Happo, can you rebase on latest `master`? I've seen failures when the branch is out of sync before.

williaster:
@RobinClowers in looking at this more closely, the happo diff that is failing is for the `DateTimeSelect` story `Fallback to today's date if value is invalid`, the diff shows a change from yesterday's date to today's date ... so it seems like it would make sense.

I'm a little confused because stories should be deterministic to avoid things like this, and I'd expect for this to have come up before if it changes daily 🤔 @alecklandgraf do you have any thoughts? 

For now I'll approve the diff and we can merge 👍

Octember:
Nice catch @williaster 

This is the story that was failing, and I agree with your argument https://github.com/airbnb/lunar/blob/master/packages/core/src/components/DateTimeSelect/story.tsx#L173-L186

Seems like it falls back on current time

🚢

PR #377, opened 2020-05-29
new(Popover): Add Popover component

felixcodes:
"to: @williaster @hayes @alecklandgraf

## Description

Popover is essentially a tooltip component with the ability to persist the popup by mouse-entering the popup portion of the tooltip. This functionality is achieved by adding a delay timer and setTimeout (via `delaySetPopupVisible`) to ensure that the popup persists long enough for a mouse to move from tooltip trigger to popup div. When a mouse enters the popup div, the timer is cleared, and the popup can persist until the mouse leaves the popup. 

I add two new props: `mouseEnterDelay`, and `mouseLeaveDelay` so a client can specify the amount of time they would like the cursor to persist on the trigger before displaying a popup, and the minimum amount of time a popup can persist after the mouse leaves the trigger or popup. These props are demoed in the following story: `/story/core-popover--custom-delay-popover`

## Motivation and Context

Lunar’s current tooltip is intended for use by items that are not important enough to be copied or interacted with. However, several internal tools at Airbnb now require the ability to click links inside a popup, and copy-and-paste text in the popup, leading to the creation of this popover component. 
 
The popover is expected to function like the lunar tooltip (ex. layout, positioning), but with the ability to interact with content in the overlay (copy text, click links, etc.). 

We need be able to open the popover by `mouse-enter`ing the trigger, and close the popover by `mouse-leave`ing either the popup or the trigger.  

## Testing

Added enzyme tests to ensure parity with the Tooltip component, which Popover shares most of its source code with. Added additional tests to cover new props and behavior of Popover, such as `mouseEnterDelay` and `mouseLeaveDelay`. Tested locally in storybook.

## Screenshots

image

image

image

## Checklist


alecklandgraf:
Your popover tests are failing, also please don't use `toMatchSnapshot`

packages/core/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+import Portal from '../Portal';
+const EMPTY_TARGET_RECT: ClientRect = {
+  bottom: 0,
+  height: 0,
+  left: 0,
+  right: 0,
+  top: 0,
+  width: 0,
+};
+export type PopoverProps = {
```
stefhatcher, line 16: Are any of these props shared with Tooltip?

packages/core/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+  /** True to toggle popover display on click.  */
+  toggleOnClick?: boolean;
+  /** True to add a dotted bottom border. */
+  underlined?: boolean;
+  /** Manually override calculated vertical align */
+  verticalAlign?: 'above' | 'below';
+  /** Width of the popover in units. */
+  width?: number;
+};
+export type PopoverState = {
```
stefhatcher, line 46: please use new lines between blocks of code (between type blocks, between functions, etc.), so it's easier to read and not squished.

packages/core/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,308 @@
...
+  above: boolean;
+  align: 'left' | 'center' | 'right';
+};
+export type StyleStruct = {
+  center: string | number;
+  right: string | number;
+};
+/** A simple popover component */
+export class Popover extends React.Component<PopoverProps & WithStylesProps, PopoverState> {
+  static inverted: boolean = false;
```
stefhatcher, line 63: curious why this needed to be added as static? inverted is falsey (false or undefined) by default.

stefhatcher: actually, looking over this and Tooltip, they share _a _lot__ of logic. did you explore using Tooltip under the hood and adding private props to Tooltip instead to make this work?

https://github.com/airbnb/lunar/blob/master/packages/core/src/components/Tooltip/index.tsx

stefhatcher: Curious, why not add these props to Tooltip instead? Why a separate component? Visually they look the same, just added functionality.

felixcodes: The popup portion of the Tooltip component also has a giant <Overlay> wrapping the <span> that prevents copy-paste and clicking inside the popup, whereas this Popover does not. Otherwise, I agree -- this component is basically Tooltip with a few new props, a timer, and no overlay, and it should be a lot simpler. 

Which direction would you recommend with this context? Importing / extending Tooltip props? Modifying Tooltip? Or continuing with the separate Popover component?

stefhatcher:
Yeah, but since then, a bunch of functionality has been added to Tooltip, including options to keep it open and a bunch of callbacks. It’s way more complex now and this new functionality would fit in better. Or at least using Tooltip under the hood so all that logic to position and render it isn’t duplicated.

stefanie (mobile)

stefhatcher:
Maybe a refactoring out a BaseTooltip that they both use (under the private directory with other Base components). Then Tooltip and Popovers special stuff is applied to them directly. Or using @ignore and @private props on Tooltip to achieve what’s needed for Popover and Popover uses Tooltip under the hood.

stefanie (mobile)

milesj:
This should already be possible with `Tooltip` and these 2 props: https://github.com/airbnb/lunar/blob/master/packages/core/src/components/Tooltip/index.tsx#L37 (without the delay logic).

felixcodes:
Hey @milesj ! Sorry I missed your reply earlier.

We can use those two props (`remainOnMouseDown` & `toggleOnClick`) to persist the popups, but we won't be able to interact with the info in the popup since it's wrapped in an overlay. 

There's also an edge case when wrapping a Link component with the current tooltip: if we use the `remainOnMouseDown` prop or `toggleOnClick`, we'd be forced to click on the link to open the popup, which would trigger the link whether the user wants to or not. 

I think it would be better to factor out a BaseTooltip class like Stefanie suggested.

PR #382, opened 2020-06-25
new(Tooltip): Add `popover` prop

felixcodes:

## Description

A popover is essentially a tooltip component with the ability to persist the popup by mouse-entering the popup portion of the tooltip. 

This functionality is achieved by adding a delay timer and setTimeout (via `delaySetPopupVisible`) to ensure that the popup persists long enough for a mouse to move from tooltip trigger to popup div when the `popover` prop is selected. When a mouse enters the popup div, the timer is cleared, and the popup can persist until the mouse leaves the popup.
## Motivation and Context

Several internal tools at Airbnb now require the ability to click links inside a popup, and copy-and-paste text in the popup. However, without this prop, it is not possible to interact with the contents of a popup in the Lunar tooltip

(`remainOnMouseDown` & `toggleOnClick`) props _can_ be used to persist the popups, but we won't be able to interact with the info in the popup since `renderPopup` previously only rendered everything in an Overlay. Additionally, we can't use these props the way we'd like when wrapping a link component with a Tooltip that has these props -- you'd click the link when triggering these props. The timeout method via `delaySetPopupVisible` solves this problem.

## Testing
Added Jest tests to cover the expected open and closing behavior of a popover. Tested locally in storybook.
## Screenshots

image

image

## Checklist

### DIFF
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -17,7 +17,7 @@ const EMPTY_TARGET_RECT: ClientRect = {
   width: 0,
 };
 
-const MOUSE_LEAVE_DELAY = 100; // 100ms
+const MOUSE_LEAVE_DELAY_MS = 100;
 
 export type TooltipProps = {
   /** Accessibility label. If not specified, all tooltip content is duplicated, rendered in an off-screen element with a separate layer. */
@@ -100,8 +100,7 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
 
   rafHandle: number = 0;
 
-  // Only used when Popover is enabled
-  delayTimeoutId = 0;
+  popoverDelayTimeoutId = 0;
 
   static getDerivedStateFromProps({ disabled }: TooltipProps) {
     if (disabled) {
@@ -226,7 +225,7 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
   private handleMouseLeave = () => {
     if (!this.props.toggleOnClick) {
       if (this.props.popover) {
-        this.delayedSetPopoverVisible(false, MOUSE_LEAVE_DELAY);
+        this.delayedSetPopoverVisible(false, MOUSE_LEAVE_DELAY_MS);
       } else {
         this.handleClose();
       }
@@ -246,13 +245,13 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
   }
 
   clearDelayTimer() {
-    clearTimeout(this.delayTimeoutId);
+    clearTimeout(this.popoverDelayTimeoutId);
   }
 
   delayedSetPopoverVisible(open: boolean, delayMs: number = 0) {
     this.clearDelayTimer();
     if (delayMs) {
-      this.delayTimeoutId = window.setTimeout(() => {
+      this.popoverDelayTimeoutId = window.setTimeout(() => {
         this.setPopoverVisible(open);
       }, delayMs);
     } else {
@@ -265,7 +264,7 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
   };
 
   handlePopoverMouseLeave = () => {
-    this.delayedSetPopoverVisible(false, MOUSE_LEAVE_DELAY);
+    this.delayedSetPopoverVisible(false, MOUSE_LEAVE_DELAY_MS);
   };
 
   private renderPopUp() {
@@ -327,7 +326,7 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
     );
 
     if (popover) {
-      return open && <div style={{ position: 'absolute', zIndex: 1 }}>{popupContent}</div>;
+      return open && <div className={cx(styles.popover)}>{popupContent}</div>;
     }
 
     return (
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -17,8 +17,7 @@ const EMPTY_TARGET_RECT: ClientRect = {
   width: 0,
 };
 
-const MOUSE_ENTER_DELAY = 0;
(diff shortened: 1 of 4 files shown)


packages/core/src/components/Tooltip/index.tsx
```diff
@@ -95,6 +101,9 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
 
   rafHandle: number = 0;
 
+  // Only used when Popover is enabled
+  delayTimer: number | null = null;
```
williaster, line 105: you can simplify this logic by setting `delayTimeoutId = 0;` because they are always non-zero. then you don't have to do checks. (this is also consistent with how I initialized `rafHandle` above)

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -213,8 +226,52 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
...
+  }
+
+  clearDelayTimer() {
+    if (this.delayTimer) {
+      clearTimeout(this.delayTimer);
+      this.delayTimer = null;
+    }
+  }
+
+  delaySetPopupVisible(open: boolean, delayS: number) {
```
williaster, line 256: this name is a little confusing to me. what do you think about `delayedSetPopoverVisible`?  (also `popup` => `popover` to be consistent?)

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -250,22 +308,37 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
...
-          <div className={cx(styles.content, invert && styles.content_inverted)}>
-            <Text inverted={invert}>{content}</Text>
-          </div>
+          <Text inverted={invert}>{content}</Text>
         </div>
+      </div>
+    );
+
+    if (popover) {
+      return open && <div style={{ position: 'absolute', zIndex: 1 }}>{popupContent}</div>;
```
williaster, line 336: ~these should be in `styles` ideally, not inline~

Edit: It would be one more addition to this PR, but I think we should update `Overlay` to [optionally apply the `userSelect: 'none'`](https://github.com/airbnb/lunar/blob/master/packages/core/src/components/Overlay/Portal/styles.ts#L13) style with a prop. 
Reasons:
- Overlay handles a lot of logic, like updating the position upon scroll (with complex scroll listeners), etc. which I'm not sure we handle here (?)
- This would enable us to use `onClose` irrespective of whether `popover=true/false`

I played with it in the storybook, and it doesn't seem like it's necessary. But I think better to add it as a prop to enable it so it's not a breaking change.

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -213,8 +226,52 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
...
+      this.delayTimer = window.setTimeout(() => {
+        this.setPopupVisible(open);
+        this.clearDelayTimer();
+      }, delay);
+    } else {
+      this.setPopupVisible(open);
     }
+  }
+
+  handlePopupMouseEnter = () => {
```
williaster, line 269: `popup` => `popover` to be consistent?

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -194,7 +203,11 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
 
   private handleMouseEnter = () => {
     if (!this.props.toggleOnClick) {
-      this.handleOpen();
+      if (this.props.popover) {
+        this.delaySetPopupVisible(true, MOUSE_ENTER_DELAY);
```
williaster, line 207: I understand the mouse out delay, but I'm not sure I follow why we need the delay in visibility. can we remove this? esp since it is `0`?

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -213,8 +226,52 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
...
+    }
+  }
+
+  delaySetPopupVisible(open: boolean, delayS: number) {
+    const delay = delayS * 1000;
+    this.clearDelayTimer();
+    if (delay) {
+      this.delayTimer = window.setTimeout(() => {
+        this.setPopupVisible(open);
+        this.clearDelayTimer();
```
williaster, line 262: I don't think you need to clear this because it will have executed by the time this runs.

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -17,6 +17,9 @@ const EMPTY_TARGET_RECT: ClientRect = {
   width: 0,
 };
 
+const MOUSE_ENTER_DELAY = 0;
+const MOUSE_LEAVE_DELAY = 0.1;
```
williaster, line 21: is there a reason to make these seconds? typically I've seen these defined as `ms` because that's what you have to pass to timeout apis.

packages/core/src/components/Tooltip/index.tsx
williaster, line 256: this could become `delayMs` if they are updated.

packages/core/src/components/Tooltip/index.tsx
williaster: if we don't use `Overlay`, we should move these styles into the stylesheet.

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -34,6 +36,8 @@ export type TooltipProps = {
   onClose?: () => void;
   /** Callback fired when the tooltip is shown. */
   onShow?: () => void;
+  /** True to enable interactive popover functionality */
```
williaster, line 39: super nit but all of these should be full sentences / end in `.` 😬

felixcodes: missing `.`? what missing `.`? (i'll add the period in a later tooltip addition, don't want to wait for CI again 😭 

image

PR #384, opened 2020-07-01
enabling customization of inner modal content

dquang-tran:

## Description

styleSheet prop is passed from the parent component but is currently not being used.
This pr propagates styleSheet to both InnerModal and InnerContent

## Motivation and Context

To customize the styles of modal

### DIFF
--- a/packages/core/src/components/Modal/private/Inner.tsx
+++ b/packages/core/src/components/Modal/private/Inner.tsx
@@ -1,11 +1,11 @@
 import React from 'react';
-import withStyles, { WithStylesProps } from '../../../composers/withStyles';
+import useStyles, { StyleSheet } from '../../../hooks/useStyles';
 import FocusTrap from '../../FocusTrap';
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Modal/private/Inner.tsx
```diff
@@ -78,21 +81,23 @@ export class ModalInner extends React.Component<ModalInnerProps & WithStylesProp
...
-      styles,
       subtitle,
       title,
       topBar,
       topBarCentered,
+      styleSheet,
+      innerContentStyleSheet,
     } = this.props;
 
+    const [styles, cx] = useStyles(styleSheet ?? styleSheetInner);
```
williaster, line 99: you can't use a hook in a class component right?

williaster:
Also sorry, CI is failing because it won't work for forks. needs to be a branch on `airbnb`.

packages/core/src/components/Modal/private/Inner.tsx
milesj: No you cannot.

PR #385, opened 2020-07-01
update(tooltip): Make inverted text readable in dark mode

Xapphire13:

## Description
When in dark mode, the inverted tooltip text isn't readable. This change fixes that, and also improves the accessibility of the colors for light mode.

In order to make this fix, ~I needed to have theme mode aware styles or use some other color from the theme; there were no applicable analogs, so I went the route of exposing the current theme mode in the theme.~ I added a new `baseInverse` color value to the theme.

## Motivation and Context
The colors currently used for the inverted tooltip aren't accessible, causing difficulty for viewers of all eyesights

## Testing
- Tested with storybook

## Screenshots
**Light Mode**
Before | After
-|-
image | image
Notice how the transparency decreases the contrast, making it harder to read |

**Dark Mode**
Before | After
-|-
image | image
Notice how the text is almost completely unreadable |

## Checklist

### DIFF
--- a/packages/core/src/components/Tooltip/styles.ts
+++ b/packages/core/src/components/Tooltip/styles.ts
@@ -1,6 +1,6 @@
 import { StyleSheet } from '../../hooks/useStyles';
 
-export const styleSheetTooltip: StyleSheet = ({ unit, color, pattern, ui }) => ({
+export const styleSheetTooltip: StyleSheet = ({ unit, color, pattern, ui, mode }) => ({
   container: {
     display: 'inline-block',
   },
@@ -55,7 +55,7 @@ export const styleSheetTooltip: StyleSheet = ({ unit, color, pattern, ui }) => (
   },
 
   content_inverted: {
-    backgroundColor: color.accent.blackout,
+    backgroundColor: mode === "light" ? "#000" : "#FFF",
   },
 
   popover: {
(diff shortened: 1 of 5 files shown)


kenchendesign:
Good touch! I like that you are removing the transparency but I'd suggest using the same dark gray as in the primary text since we avoid absolute black in the UI.

alecklandgraf:
The happo diff is odd 
image

packages/core/src/components/Tooltip/styles.ts
```diff
@@ -55,7 +55,10 @@ export const styleSheetTooltip: StyleSheet = ({ unit, color, pattern, ui }) => (
   },
 
   content_inverted: {
-    backgroundColor: color.accent.blackout,
+    backgroundColor:
+      mode === 'light'
+        ? '#181818' // Dark mode base color
+        : '#FFF', // Light mode base color
```
alecklandgraf, line 61: do these exist on `color`?

williaster:
@alecklandgraf there's a component that falls back to today's date which means that it's happo is non-deterministic. we should fix it, for now I've been approving the diff.

Xapphire13:
@williaster, that's something I also considered. I'll switch to that instead =]

packages/core/src/components/Tooltip/styles.ts
Xapphire13: Only on the opposite theme 🤔 . Should be solved by using the new `baseInverse`

Xapphire13:
@williaster @alecklandgraf, Your comments should be addressed now ^_^

Xapphire13:
Abandoning in favor of https://github.com/airbnb/lunar/pull/386

PR #390, opened 2020-07-17
fix(Autocomplete): prefix and suffix style

moyus:

## Description

`BaseInput` in `Autocomplete` doesn't receive `hasPrefix` and `hasSuffix` props before, which would cause style bug with `<Prefix />` and `<Suffix />`.

The `hasPrefix` and `hasSuffix` condition check is copied from `partitionFieldProps()` function.

## Motivation and Context

Fix style bug.

## Testing

Looks good in storybook.

## Screenshots

Before:
Screen Shot 2020-07-17 at 4 00 06 PM

After:
Screen Shot 2020-07-17 at 3 57 52 PM

## Checklist

### DIFF
--- a/packages/core/src/components/Autocomplete/index.tsx
+++ b/packages/core/src/components/Autocomplete/index.tsx
@@ -445,8 +445,8 @@ export default class Autocomplete<T extends Item = Item> extends React.Component
     return items;
   }
 
-  getInputProps(props: AutocompleteProps<T>) {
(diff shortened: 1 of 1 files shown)


williaster:
@moyus looks like this breaks `getInputProps` tests as several of the expected props are wrong. Would another approach be to update `partitionFieldProps` such that `prefix` and `suffix` are included in the `inputProps` it generates?

moyus:
@williaster fixed, please take a look.

PR #391, opened 2020-07-23
new(Overlay): Add `enableMouseInteraction` prop

felixcodes:

## Description

This PR adds the `enableMouseInteraction` prop to `Overlay` and `Overlay/Portal` to enable mouse events and mouse interaction in Overlays.  In the Overlay portal stylesheet, we extract the `container` class' `userSelect: 'none'` out into a `noUserSelect` class, and introduce the ~`hasPointerEvents`~ `withPointerEvents` class to explicitly enable pointer events in Overlay. These two classes are composed in such a way that there are no breaking changes. 

## Motivation and Context

Overlay's portal implementation is useful for Tooltip and Popover popups, which need to escape their parent divs. The current Popover implementation renders the popup in a div with a set zIndex, which does not let the popup escape a DataTable row, while a plain Tooltip's `<Overlay></Overlay>` popup is able to escape the DataTable as intended.

To allow Popover to use `<Overlay>` for its Portal rendering and positioning logic, we added the ability to enable pointer events in Overlay so that Popover's mouse event based timeout logic and copy-paste functionality could work when implemented with Overlay.

## Testing

Tested in storybook. 

## Screenshots

An overlay without `enableMouseInteraction` prop:
overlay-no-prop

An overlay with `enableMouseInteraction` prop (mouse can interact with content in the overlay):
overlay-with-prop

## Checklist

### DIFF
--- a/packages/core/src/components/Overlay/Portal/styles.ts
+++ b/packages/core/src/components/Overlay/Portal/styles.ts
@@ -12,7 +12,7 @@ export const styleSheetPortal: StyleSheet = ({ color }) => ({
     overflow: 'auto',
   },
 
-  hasPointerEvents: {
+  withPointerEvents: {
     pointerEvents: 'all',
   },
 
--- a/packages/core/src/components/Overlay/Portal/styles.ts
+++ b/packages/core/src/components/Overlay/Portal/styles.ts
@@ -10,12 +10,19 @@ export const styleSheetPortal: StyleSheet = ({ color }) => ({
     right: 0,
     bottom: 0,
     overflow: 'auto',
-    userSelect: 'none',
+  },
+
+  hasPointerEvents: {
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Overlay/Portal/styles.ts
```diff
@@ -10,14 +10,21 @@ export const styleSheetPortal: StyleSheet = ({ color }) => ({
     right: 0,
     bottom: 0,
     overflow: 'auto',
-    userSelect: 'none',
+  },
+
+  hasPointerEvents: {
```
williaster, line 15: wdyt about `withPointerEvents` instead?

felixcodes: sgtm, will make that change and push

PR #392, opened 2020-07-23
update(Tooltip): Render Popover's popup in Overlay component

felixcodes:

## Description
This PR refactors Popover to render inside an `<Overlay>` component, just like the plain tooltip. It also moves the popup `mouseEnter` and `mouseLeave` handlers up to the popup content's outermost div.

Note: We'll want to land #391 first, and then update the target of this PR to master. I set the base of this PR to the Overlay PR to facilitate code review.

## Motivation and Context
The original Popover component rendered in a simple div with a style of `zIndex:1`. This was naive on my part for several reasons: 
1. The Popover would need a much larger zIndex to escape divs with zIndexes potentially much larger than 1. 
2. In our Popover-inside-DataTable use case, the Popover was unable to escape a DataTable row, while a regular Tooltip, which rendered in Overlay's portal, was able to. 

## Testing

Tested with existing unit tests and new story book story.
## Screenshots
popover-overlay-works

## Checklist

### DIFF
--- a/packages/core/src/components/Tooltip/index.tsx
+++ b/packages/core/src/components/Tooltip/index.tsx
@@ -314,23 +314,17 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
           marginTop: above ? -(tooltipHeight + targetRect.height + distance) : distance,
           textAlign: align,
         })}
+        onMouseEnter={handleMouseEnter}
+        onMouseLeave={handleMouseLeave}
       >
-        <div
-          className={cx(styles.content, invert && styles.content_inverted)}
-          onMouseEnter={handleMouseEnter}
-          onMouseLeave={handleMouseLeave}
-        >
+        <div className={cx(styles.content, invert && styles.content_inverted)}>
           <Text inverted={invert}>{content}</Text>
         </div>
       </div>
     );
 
-    if (popover) {
-      return open && <div className={cx(styles.popover)}>{popupContent}</div>;
-    }
-
     return (
-      <Overlay noBackground open={open} onClose={this.handleClose}>
+      <Overlay noBackground enableMouseInteraction={popover} open={open} onClose={this.handleClose}>
         {popupContent}
       </Overlay>
     );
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Tooltip/index.tsx
```diff
@@ -314,23 +314,17 @@ export class Tooltip extends React.Component<TooltipProps & WithStylesProps, Too
           marginTop: above ? -(tooltipHeight + targetRect.height + distance) : distance,
           textAlign: align,
         })}
+        onMouseEnter={handleMouseEnter}
```
williaster, line 317: can you elaborate on why this change was needed? Seems fine, wanted to check that the `margin` areas didn't trigger the event listeners tho (like to the top/left of the actual content)

packages/core/src/components/Tooltip/index.tsx
```diff
@@ -28,23 +28,23 @@ export type TooltipProps = {
...
+  /** Manually override calculated horizontal align. */
   horizontalAlign?: 'center' | 'left' | 'right';
   /** True to use a light background with dark text. */
   inverted?: boolean;
   /** Callback fired when the tooltip is closed. */
   onClose?: () => void;
   /** Callback fired when the tooltip is shown. */
   onShow?: () => void;
-  /** True to enable interactive popover functionality */
+  /** True to enable interactive popover functionality. */
```
williaster, line 39: thanks for adding these 😄

felixcodes: np, I said I would in the last PR :)

packages/core/src/components/Tooltip/index.tsx
felixcodes: >wanted to check that the margin areas didn't trigger the event listeners tho (like to the top/left of the actual content)
I can assure this is ok with a gif (hacked in a larger margin so we can see the gif easily). 
margins-no-problem

>can you elaborate on why this change was needed? 
Yesterday Erik and I ran into a weird case where if the mouse was on the trigger while the popup was opening in the direction the mouse came from, there wouldn't be a continuity between  mouse enter trigger -> mouse leave trigger (because of the popup animation) -> mouse entering popup -> mouse leaving popup (when animations passes the cursor) -> mouse back on trigger. 

Oddly enough, I don't seem to be encountering that today if I move the mouse handlers back down to the content div. But the first screenshot shows margins weren't causing an issue with the current implementation, so I think the handlers should be fine on either div.

continuity

PR #396, opened 2020-07-29
new(Card): Adding two props 'noShadow' and 'selected' for Card

monicapharm:

## Description
Add two new props for the `Card` component. 
1.  `noShadow` that hides the shadow of the card.
2. `selected` that shows the border when selected.

## Motivation and Context
**Background:** 
Some discussions are here https://git.musta.ch/airbnb/solar/issues/978, the design team call out for adding a new variation '`noShadow`' for Card to support some features like the wrapper of checkboxes.

And in the long run, looks like the Lunar team also consider to refactor `clickable button checkboxes` to just being wrapped with `card`, so also adding the `selected` state here in this PR. 

**Design:** [figma](https://www.figma.com/file/JRX1IRigC6oxce2nyFqAZM/Nova-Guides---V2-Features?node-id=1016%3A64864)
image

## Testing
Local storybook and unit tests. 

## Screenshots
When there is **noShadow**: 
image

When **selected**: 
image

## Checklist

### DIFF
--- a/packages/core/src/components/Card/index.tsx
+++ b/packages/core/src/components/Card/index.tsx
@@ -26,7 +26,8 @@ export default function Card({ children, noShadow, overflow, selected, styleShee
 
   return (
     <div
-      className={cx(styles.card,
+      className={cx(
+        styles.card,
         overflow && styles.card_overflow,
         noShadow && styles.card_noShadow,
         selected && styles.card_selected,
--- a/packages/core/src/components/Card/index.tsx
+++ b/packages/core/src/components/Card/index.tsx
@@ -26,8 +26,7 @@ export default function Card({ children, noShadow, overflow, selected, styleShee
 
   return (
     <div
-      className={cx(
-        !noShadow && styles.card,
+      className={cx(styles.card,
         overflow && styles.card_overflow,
         noShadow && styles.card_noShadow,
         selected && styles.card_selected,
--- a/packages/core/src/components/Card/index.tsx
+++ b/packages/core/src/components/Card/index.tsx
@@ -8,8 +8,12 @@ export { Content };
 export type CardProps = {
   /** List of `Content`s blocks to contain content. */
   children: NonNullable<React.ReactNode>;
+  /** Hide shadow. */
+  noShadow?: boolean;
   /** Set overflow to be visible. */
   overflow?: boolean;
+  /** Whether the card is selected or not. */
+  selected?: boolean;
   /** Custom style sheet. */
   styleSheet?: StyleSheet;
 };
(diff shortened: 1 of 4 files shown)


monicapharm:
2 added in Happo report:

image

monicapharm:
@williaster @alecklandgraf PTAL when you got a chance. : )

packages/core/src/components/Card/styles.ts
```diff
@@ -10,6 +10,14 @@ export const styleSheetCard: StyleSheet = ({ color, pattern }) => ({
   card_overflow: {
     overflow: 'visible',
   },
+
+  card_noShadow: {
+    border: ui.border,
+  },
+
+  card_selected: {
+    border: `2px solid ${color.core.primary[3]}`,
```
williaster, line 19: can you use [`ui.borderThick`](https://github.com/airbnb/lunar/blob/master/packages/core/src/themes/buildTheme.ts#L56) here instead?

packages/core/src/components/Card/index.tsx
```diff
@@ -8,17 +8,32 @@ export { Content };
...
  */
-export default function Card({ children, overflow, styleSheet }: CardProps) {
+export default function Card({ children, noShadow, overflow, selected, styleSheet }: CardProps) {
   const [styles, cx] = useStyles(styleSheet ?? styleSheetCard);
 
-  return <div className={cx(styles.card, overflow && styles.card_overflow)}>{children}</div>;
+  return (
+    <div
+      className={cx(
+        !noShadow && styles.card,
```
williaster, line 30: I think the way you've written this breaks the component. `styles.card` sets 
- `pattern.box` (`border/boxShadow/borderRadius`) 
- `backgroundColor`
- `overflow`

and your `styles.card_noShadow` only applies `border` and ignores the rest. it seems like we could **always** apply `styles.card`, and re-define `card_noShadow` to only override `boxShadow`?

packages/core/src/components/Card/styles.ts
monicapharm: Sure, will do.

packages/core/src/components/Card/index.tsx
monicapharm: Good catch! :)

packages/core/src/components/Card/index.tsx
```diff
@@ -8,17 +8,31 @@ export { Content };
...
  * An abstract layout to use as a base for cards.
  */
-export default function Card({ children, overflow, styleSheet }: CardProps) {
+export default function Card({ children, noShadow, overflow, selected, styleSheet }: CardProps) {
   const [styles, cx] = useStyles(styleSheet ?? styleSheetCard);
 
-  return <div className={cx(styles.card, overflow && styles.card_overflow)}>{children}</div>;
+  return (
+    <div
+      className={cx(styles.card,
```
williaster, line 29: @monicapharm I think prettier is still failing in CI, I think because of this line

monicapharm:
@williaster Do you know why the pull request CI failed? looks like it's not related to my change 🤔

williaster:
happo has been flaky recently 😞  re-running them.

monicapharm:
Re-run the CI a couple of times, finally got them passed 😄  @williaster PTAL

PR #397, opened 2020-07-30
update(Link): Add `truncated` prop

elizabeth-moore:

## Description

This adds a `truncated` property to links and a `link_truncated` style, as well as a unit test. It also removes `truncated` from the `textProps` because it doesn't work.

## Motivation and Context

Although Link takes in `textProps` to pass to the Text component, `overflow` and `text-overflow` aren't inherited, so the truncation isn't applied. This removes the ability to define `truncated` in the `textProps` and adds a prop to Link so that truncating a Link with ellipsis is possible.

## Testing

I tested this locally in storyguide and added a unit test.

## Screenshots

Screen Shot 2020-07-30 at 4 43 52 PM

## Checklist

### DIFF
--- a/packages/core/src/components/Link/index.tsx
+++ b/packages/core/src/components/Link/index.tsx
@@ -23,12 +23,14 @@ export type LinkProps = ButtonOrLinkProps & {
   muted?: boolean;
   /** Decrease font size to small. */
   small?: boolean;
+  /** Truncate the link text with an ellipsis. */
+  truncated?: boolean;
   /** Bold font. */
   bold?: boolean;
   /** Custom style sheet. */
   styleSheet?: StyleSheet;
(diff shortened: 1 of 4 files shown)


williaster:
This looks great! 

One thought is that a new storybook story might useful since it will give us a happo visual diff test, which will ensure functional correctness more than a unit test which asserts on a className. Happo also seems to be glitching right now (5 diffs for dark only seem unrelated to this), so adding a story and re-running CI might fix that.

williaster:
All of the jobs passed according to [this](https://github.com/airbnb/lunar/pull/397/checks?check_run_id=933211918), but the `happo (light)` job isn't updating here. gonna merge so we can get this released 🎉

PR #403, opened 2020-10-09
fix(ImageViewer): Only call document.preventDefault() when dragging

apiel51:

## Description
`document.preventDefault` should only be called when `dragging` is `true`, otherwise it's impossible to highlight text and in the document

## Motivation and Context
Running into this error in Torch

## Testing
Tested locally in Storybook, added a text component above `<ImageViewer />`, tried to highlight and it didn't work. After this change, highlighting worked.

## Screenshots
No visual diff

## Checklist


williaster (approved):
nice

PR #406, opened 2020-10-23
fix(Autocomplete): isItemSelected should check against input value

ktmud:

## Description

TODO

## Testing

TODO

## Checklist

### DIFF
--- a/packages/core/src/components/Autocomplete/index.tsx
+++ b/packages/core/src/components/Autocomplete/index.tsx
@@ -643,7 +643,7 @@ export default class Autocomplete<T extends Item = Item> extends React.Component
     return this.getFilteredItems(this.state).map((item: T, index) => {
       const value = this.props.getItemValue!(item);
       const selected = this.props.isItemSelected
-        ? this.props.isItemSelected(item, value)
+        ? this.props.isItemSelected(item, this.state.value)
         : value === this.state.value;
       const props: React.HTMLAttributes<HTMLDivElement> = {};
 


packages/core/src/components/Autocomplete/index.tsx
```diff
@@ -643,7 +643,7 @@ export default class Autocomplete<T extends Item = Item> extends React.Component
     return this.getFilteredItems(this.state).map((item: T, index) => {
       const value = this.props.getItemValue!(item);
       const selected = this.props.isItemSelected
-        ? this.props.isItemSelected(item, value)
+        ? this.props.isItemSelected(item, this.state.value)
```
williaster, line 646: @ktmud I'm not sure I agree this is a bug. it looks like the api is simply passing the item value as the second argument, not the autocomplete value.

This would be a breaking change if someone had logic that assumed the second argument was the item value.

ktmud: Currently there is not way to compare item value against `state.value` since `isItemSelected` has a default value of `() => false`, which you cannot override by passing `undefined`, the only other compatible value according to typing definition.

It would be weird (and probably not advisable) to check `isItemSelected` without checking it against the real selected value. Plus users shouldn't need item values, either, because they can easily get it from `getItemValue`.

williaster: > Plus users shouldn't need item values, either, because they can easily get it from getItemValue.
agree with this for sure

> Currently there is not way to compare item `value` against `state.value`
can you give more context to when this is an issue? wouldn't you be able to get `value` from `onChange` callbacks? or in the case of a controlled component the user would be passing the `value` themselves?

ktmud: This function (`isItemSelected`) is used when rendering the menu items. I wanted to render selected values differently in the menu. Currently it's not possible to do that.

williaster (approved):
Awesome, thanks for entertaining the non-breaking change idea 🙏

PR #408, opened 2020-11-01
update(Spacing): Add 0.25 to SpacingRange

apiel51:

## Description
Title.

## Motivation and Context
Needed for a Torch design.

## Testing
N/A

## Screenshots
N/A

## Checklist

### DIFF
--- a/packages/core/src/components/Spacing/index.tsx
+++ b/packages/core/src/components/Spacing/index.tsx
@@ -4,6 +4,7 @@ import { styleSheetSpacing, cleanClassName } from './styles';
 
 export type SpacingRange =
   | 0
+  | 0.25
   | 0.5
   | 1
   | 1.5


packages/core/src/components/Spacing/index.tsx
```diff
@@ -4,6 +4,7 @@ import { styleSheetSpacing, cleanClassName } from './styles';
 
 export type SpacingRange =
   | 0
+  | 0.25
```
williaster, line 7: @apiel51 I think if we are going to support `0.25` it needs to be throughout the range, not a one-off

apiel51: @williaster  Yeah, I had thought about that as well. I figured that supporting `0.25` throughout the range was a little overkill, and it deemed only adding `0.25` permissible since we're only adding `0.25` itself and not something like `6.75`. But happy to include this granularity across the entire range if you think that's better!

williaster: I agree that `0.25` increments throughout the range is probably overkill, but that would be better than being inconsistent. In instances where we've needed `< 4px` of a margin, we've just added custom css which you could also consider 😄

ktmud: Why can't we just allow arbitrary number? I understand that enforces some visual consistency but doesn't having a unit system already suffice?

williaster: Theoretically that's fine but I think there was intentional optimization of the style sheet, to [enumerate the possible values ahead of time](https://github.com/airbnb/lunar/blob/8ae2f394c24928a739104baaac3c24a7e7f86d27/packages/core/src/components/Spacing/styles.ts#L11).

I think actually the PR currently doesn't work because `0.25` doesn't have a matching style. (screenshots would help test this 😄 )

ktmud: Can we then maybe just use inline style? It has no performance concerns. 

In my project, I have this very simple `InlineStyle` component, which arguably is the most useful component I've created for Lunar styling...

```ts
import React, { CSSProperties } from 'react';
import useTheme from '@airbnb/lunar/lib/hooks/useTheme';

export type InlineStyleProps = {
  tag?: keyof JSX.IntrinsicElements;
  children?: React.ReactNode;
  inline?: boolean;
} & CSSProperties;

const dimensionProps: (keyof CSSProperties)[] = [
  'width',
  'height',
  'margin',
  'padding',
  'left',
  'right',
  'bottom',
  'top',
  'marginTop',
  'marginLeft',
  'marginRight',
  'marginBottom',
  'paddingTop',
  'paddingLeft',
  'paddingRight',
  'paddingBottom',
];

export default function InlineStyle({
  tag = 'div',
  children,
  inline = false,
  ...style
}: InlineStyleProps) {
  const Tag = tag as 'div';
  const { unit } = useTheme();

  /**
   * Process inline styles with steps:
   *   1. times numeric values of dimension properties by `unit`
   *   2. .. maybe more
   */
  function processInlineStyles(styles: Partial<CSSProperties>) {
    const updatedStyle = { ...styles };
    dimensionProps.forEach(prop => {
      if (prop in updatedStyle && typeof updatedStyle[prop] === 'number') {
        // coerce a limited type set to appease TypeScript
        updatedStyle[prop as 'top'] = (updatedStyle[prop] as number) * unit;
      }
    });
    return updatedStyle;
  }

  return (
    <Tag
      style={{
        display: inline ? 'inline-block' : 'block',
        ...processInlineStyles(style),
      }}
    >
      {children}
    </Tag>
  );
}
```

apiel51: @williaster Yeah - I think you're right about using custom styles. Not sure if we've had to use < 4px before now, so maybe can re-evaluate if this becomes a more common use case. Also didn't know about the style optimization for spacing, nice catch!

PR #412, opened 2021-03-15
update(Modal): Add optional zIndex to props

nilgradisnik:

## Description

Allow for optional `zIndex` to be passed to Modal.

## Motivation and Context

We want to override z-index for Modals that are in a Lighbox component where a recent change increased the z-index https://github.com/airbnb/lunar/pull/411. This allows to bring such Modals to the front.

## Testing

Tested in storybook

## Screenshots

Screen Shot 2021-03-15 at 10 16 20 AM

## Checklist

### DIFF
--- a/packages/core/src/components/Modal/index.tsx
+++ b/packages/core/src/components/Modal/index.tsx
@@ -46,16 +46,8 @@ export default function Modal({
   return (
     <Portal>
       <div className={cx(containerZIndex, styles.container)}>
-        <div
-          role='presentation'
-          className={cx(styles.wrapper)}
-          onKeyUp={handleKeyUp}
-        >
-          <ModalInner
-            {...props}
-            styleSheet={innerStyleSheet}
-            onClose={onClose}
-          />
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Modal/index.tsx
```diff
@@ -4,14 +4,23 @@ import Portal from '../Portal';
 import ModalInner, { ModalInnerProps } from './private/Inner';
 import { ESCAPE } from '../../keys';
 import { styleSheetModal } from './styles';
+import { Z_INDEX_MODAL } from '../../constants';
 
 export type ModalProps = ModalInnerProps & {
   /** Custom style sheet. */
   innerStyleSheet?: StyleSheet;
+  /** Z-index of the modal. */
+  zIndex?: number | 'auto';
```
alecklandgraf, line 13: what's the difference between `auto` and `Z_INDEX_MODAL` here?

nilgradisnik: This is saying `zIndex` can be either a number or a string `auto`. This is a typescript type union, not a default value.

alecklandgraf (approved):

PR #417, opened 2021-08-07
fix(Autocomplete): Set items only for valid promise

pengchongfu:
to: @alecklandgraf

## Description

Set items only for valid promise.
If these are multiple promises, the last promise would set items. This could confuse user.

## Motivation and Context

If user inputs `a`, and the first promise is in pending.
And then user inputs more character `abc`, and the second promise is in pending.
If the second promise resolves first, the `items` would be set to the result of the first promise. It would confuse user.

## Testing

Unit tests.

## Screenshots

N/A

## Checklist

### DIFF
--- a/packages/core/src/components/Autocomplete/index.tsx
+++ b/packages/core/src/components/Autocomplete/index.tsx
@@ -536,6 +536,10 @@ export default class Autocomplete<T extends Item = Item> extends React.Component
           loading: false,
         };
 
+        if (input !== this.state.value) {
+          delete nextState['items'];
+        }
+
         if (callback) {
           this.setState(nextState, callback);
         } else {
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Autocomplete/index.tsx
```diff
@@ -536,6 +536,10 @@ export default class Autocomplete<T extends Item = Item> extends React.Component
           loading: false,
         };
 
+        if (input !== this.state.value) {
+          delete nextState['items'];
+        }
```
alecklandgraf, line 541: Were you able to functionally test this? `setState` is async and batches requests, so it's possible to be out of sync, you could shadow the value of `this.state.value` to a class property to be 100% sure that `input` is current. 

I think worst case is `this.state.value` falls out of sync is you'd have the same behavior you're fixing, and react would have to be in a bad state to get that far down on update state.

alecklandgraf (approved):

